PR #706, opened 2025-07-16
feat: sup 13089 test fee and ledger config changes

balanced-tree:
This PR adds two tests for fee config changes during a vault deposit/redeem lifecycle.

Test 1 has the user deposit when fee is 0%, fee is then increased to 20% and upon redemption it is asserted that 20% fee was taken.

Test 2 has the user deposit when fee is 10%, fee is then increased to 12% and upon redemption it is asserted that 12% fee was taken.

### DIFF
--- a/.github/workflows/contracts-ci.yml
+++ b/.github/workflows/contracts-ci.yml
@@ -141,8 +141,10 @@ jobs:
 
   vnet-deploy:
     needs: ["test"]
-    if:
-      (github.event_name == 'pull_request' && github.base_ref == 'main') || github.event_name == 'workflow_dispatch'
+    if: |
+      (github.event_name == 'pull_request' && github.base_ref == 'main') || 
+      (github.event_name == 'push' && github.ref == 'refs/heads/main') ||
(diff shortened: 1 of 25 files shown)


src/accounting/SuperLedgerConfiguration.sol
```diff
@@ -207,6 +207,26 @@ contract SuperLedgerConfiguration is ISuperLedgerConfiguration {
         }
     }
 
+    /// @inheritdoc ISuperLedgerConfiguration
+    function transferManagerRole(bytes32 yieldSourceOracleId, address newManager) external virtual {
+        YieldSourceOracleConfig memory config = yieldSourceOracleConfig[yieldSourceOracleId];
+        if (config.manager != msg.sender) revert NOT_MANAGER();
```
supervaulter, line 213: Maybe we can also revert with specific message if config doesn't exist. If `config.manager` == address0 for example to revert with MANAGER_NOT_SET or similar. Not very important. Just as a side note. Also the code was moved only so I think it's fine if you don't want to add this check

PR #713, opened 2025-07-17
feat: sup 13090 vault fee tests

TamaraRingas:
This PR adds isolated fee tests for 4626, 5115 and 7540 vaults.


supervaulter (approved):
I think we should also have a test with edge cases for fees like 0 to 100% allowed , multiple small redeems / large redeem and so on. Can be a separate test

supervaulter (approved):

PR #721, opened 2025-07-24
chore: removed more mocks & integration tests organization

supervaulter:


TamaraRingas (commented):
Nice so far!
I think there are a few other mocks that can still be removed though:

1. `SuperExecutor.t.sol` has `MockLedger, MockLedgerConfiguration & MockNexusFactory`
2. Ledger tests use `MockExecutorModule` instead of installing `SuperExecutor`
3. `MockEntryPoint` used in `SuperNativePaymaster` tests

PR #729, opened 2025-07-28
feat: check if deployed addresses exist and use create2 for deployment SUP-13002

0xTimepunk:
This PR:
- Fixes all notion issues (comments left there)
- Locks bytecode in a folder
- Checks if code is already deployed for that locked bytecode
- Conditionally deploys contracts that haven't been deployed yet for a given bytecode

For production it is important not to touch the locked-bytecode folder, any future upgrades should go into a new folder and new script file


script/DeployV2Base.s.sol
```diff
@@ -28,59 +41,269 @@ abstract contract DeployV2Base is Script, ConfigBase {
...
+        // Predict address first
+        address predictedAddr = DeterministicDeployerLib.computeAddress(creationCode, salt);
+
+        // Check if already deployed using assembly like Nexus
+        uint256 codeSize;
+        assembly {
+            codeSize := extcodesize(predictedAddr)
+        }
+
+        if (codeSize > 0) {
```
sujithsomraaj, line 74: could simplify this to use `predicatedAddr.code.length > 0` to be consistent with the below require statment.

script/DeployV2Base.s.sol
```diff
@@ -28,59 +41,269 @@ abstract contract DeployV2Base is Script, ConfigBase {
...
+        // Compute address
+        contractAddr = DeterministicDeployerLib.computeAddress(bytecode, args, salt);
+
+        // Check if deployed using assembly (Nexus style)
+        uint256 codeSize;
+        assembly {
+            codeSize := extcodesize(contractAddr)
+        }
+
+        isDeployed = codeSize > 0;
```
sujithsomraaj, line 174: same comment as before, assembly is not necessary here. directly use length based comparision

PR #734, opened 2025-07-31
feat: sup 13427 merkl claim hook

balanced-tree:
This PR adds MerklClaimRewardHook.sol which allows accounts to call claim() on the Merkl Distributor contract with associated unit and integration tests.

### DIFF
--- a/src/hooks/claim/merkl/MerklClaimRewardHook.sol
+++ b/src/hooks/claim/merkl/MerklClaimRewardHook.sol
@@ -74,10 +74,8 @@ contract MerklClaimRewardHook is BaseHook {
     }
 
     /// @inheritdoc ISuperHookInspector
-    function inspect(bytes calldata data) external view override returns (bytes memory) {
+    function inspect(bytes calldata data) external view override returns (bytes memory addressData) {
         (address[] memory tokens,,) = _decodeClaimParams(data);
-
-        bytes memory addressData = abi.encodePacked(distributor);
 
         uint256 length = tokens.length;
         for (uint256 i; i < length; i++) {
--- a/test/unit/hooks/claim/merkl/MerklClaimRewardsHook.t.sol
+++ b/test/unit/hooks/claim/merkl/MerklClaimRewardsHook.t.sol
@@ -84,13 +84,10 @@ contract MerklClaimRewardsHookTest is Helpers {
         bytes memory argsEncoded = hook.inspect(data);
         assertGt(argsEncoded.length, 0);
 
-        // Check that distributor is encoded correctly
-        assertEq(BytesLib.toAddress(argsEncoded, 0), distributor);
-
         // Check that tokens are encoded correctly
-        assertEq(BytesLib.toAddress(argsEncoded, 20), tokens[0]);
-        assertEq(BytesLib.toAddress(argsEncoded, 40), tokens[1]);
-        assertEq(BytesLib.toAddress(argsEncoded, 60), tokens[2]);
+        assertEq(BytesLib.toAddress(argsEncoded, 0), tokens[0]);
+        assertEq(BytesLib.toAddress(argsEncoded, 20), tokens[1]);
+        assertEq(BytesLib.toAddress(argsEncoded, 40), tokens[2]);
     }
 
     function test_CalldataDecoding() public view {


test/unit/hooks/claim/merkl/MerklClaimRewardsHook.t.sol
```diff
@@ -0,0 +1,169 @@
...
+import { IDistributor } from "../../../../../src/vendor/merkl/IDistributor.sol";
+
+// Superform
+import { Helpers } from "../../../../utils/Helpers.sol";
+import { MockERC20 } from "../../../../mocks/MockERC20.sol";
+import { BaseHook } from "../../../../../src/hooks/BaseHook.sol";
+import { BytesLib } from "../../../../../src/vendor/BytesLib.sol";
+import { ISuperHook } from "../../../../../src/interfaces/ISuperHook.sol";
+import { MerklClaimRewardHook } from "../../../../../src/hooks/claim/merkl/MerklClaimRewardHook.sol";
+
```
0xTimepunk, line 15: misses natspec

src/hooks/claim/merkl/MerklClaimRewardHook.sol
```diff
@@ -0,0 +1,206 @@
...
+    function replaceCalldataAmount(bytes memory data, uint256) external pure returns (bytes memory) {
+        return data;
+    }
+
+    /// @inheritdoc ISuperHookInspector
+    function inspect(bytes calldata data) external pure override returns (bytes memory) {
+        ClaimParams memory params = _decodeClaimParams(data);
+
+        bytes memory addressData = abi.encodePacked(params.distributor);
+        for (uint256 i = 0; i < params.users.length; i++) {
```
0xTimepunk, line 99: nit:
can remove =0 and should put the length call outside for loop
same for next loop

src/hooks/claim/merkl/MerklClaimRewardHook.sol
```diff
@@ -0,0 +1,206 @@
...
+        params.proofs = _decodeProofs(data, cursorAfterAmounts);
+    }
+
+    function _decodeUsers(bytes calldata data) internal pure returns (uint256 cursor, address[] memory users) {
+        // decode array length
+        uint256 arrayLength = BytesLib.toUint256(data, 52);
+
+        cursor = 84;
+        users = new address[](arrayLength);
+        for (uint256 i = 0; i < arrayLength; i++) {
```
0xTimepunk, line 147: Same here

src/hooks/claim/merkl/MerklClaimRewardHook.sol
```diff
@@ -0,0 +1,206 @@
...
+        internal
+        pure
+        returns (uint256 cursor, address[] memory tokens, uint256[] memory amounts)
+    {
+        uint256 arrayLength = BytesLib.toUint256(data, 52);
+
+        cursor = cursorAfterUsers;
+
+        tokens = new address[](arrayLength);
+        for (uint256 i = 0; i < arrayLength; i++) {
```
0xTimepunk, line 169: Same and in the next one too

0xTimepunk (commented):
Let's close the PR with some nit fixes

src/hooks/claim/merkl/MerklClaimRewardHook.sol
```diff
@@ -0,0 +1,196 @@
...
+    ISuperHookInspector
+} from "../../../interfaces/ISuperHook.sol";
+import { BaseHook } from "../../BaseHook.sol";
+import { HookSubTypes } from "../../../libraries/HookSubTypes.sol";
+import { HookDataDecoder } from "../../../libraries/HookDataDecoder.sol";
+
+/// @title MerklClaimRewardHook
+/// @author Superform Labs
+/// @dev data has the following structure
+/// @notice         bytes32 placeholder = bytes32(BytesLib.slice(data, 0, 32), 0);
```
supervaulter, line 24: Are we going to use this for VB? Otherwise we don't really need this here imo, bc on the 2nd position we don't have an address so cannot use decoder library anyway

src/hooks/claim/merkl/MerklClaimRewardHook.sol
```diff
@@ -0,0 +1,196 @@
...
+/// @notice         uint256 arraysLength = BytesLib.toUint256(data, 32);
+/// @notice         address[] tokens = BytesLib.slice(data, 64 + arrayLength * 20, tokensLength * 20);
```
supervaulter, line 26: Is this ok here? `arrayLength` doesn't represent the full array length? Because on this line you have `arrayLength * 20`. Same for `tokensLength`

src/hooks/claim/merkl/MerklClaimRewardHook.sol
```diff
@@ -0,0 +1,196 @@
...
+/// @notice         uint256[] amounts = BytesLib.slice(data, 64 + arrayLength * 20 + tokensLength * 20, amountsLength *
+/// 32);
+/// @notice         bytes proofBlob = BytesLib.slice(data, 64 + arrayLength * 20 + tokensLength * 20 + amountsLength *
```
supervaulter, line 29: Same length thing here

src/hooks/claim/merkl/MerklClaimRewardHook.sol
```diff
@@ -0,0 +1,196 @@
...
+/// 32, data.length - (64 + arrayLength * 20 + tokensLength * 20 + amountsLength * 32));
+contract MerklClaimRewardHook is BaseHook, ISuperHookInflowOutflow, ISuperHookOutflow, ISuperHookContextAware {
```
supervaulter, line 31: Why do we need the inheritance for InflowOutflow and Outflow?

src/hooks/claim/merkl/MerklClaimRewardHook.sol
```diff
@@ -0,0 +1,196 @@
...
+
+    struct ClaimParams {
+        address[] users;
+        address[] tokens;
+        uint256[] amounts;
+        bytes32[][] proofs;
+    }
+
+    constructor(address _distributor) BaseHook(HookType.NONACCOUNTING, HookSubTypes.CLAIM) {
+        distributor = _distributor;
```
supervaulter, line 49: Would be good to have an address0 check since it's immutable

src/hooks/claim/merkl/MerklClaimRewardHook.sol
```diff
@@ -0,0 +1,196 @@
...
+    )
+        internal
+        view
+        override
+        returns (Execution[] memory executions)
+    {
+        ClaimParams memory params;
+
+        // decode users
+        address[] memory users = _setUsersArray(account, data);
```
supervaulter, line 69: Should we check for array length ? if 0 revert ?

src/hooks/claim/merkl/MerklClaimRewardHook.sol
```diff
@@ -0,0 +1,196 @@
...
+            callData: abi.encodeCall(IDistributor.claim, (params.users, params.tokens, params.amounts, params.proofs))
+        });
+    }
+
+    /// @inheritdoc ISuperHookInflowOutflow
+    function decodeAmount(bytes memory) external pure returns (uint256) {
+        return 0;
+    }
+
+    /// @inheritdoc ISuperHookContextAware
```
supervaulter, line 88: Do we need the ISuperHookContextAware inheritance?

src/hooks/claim/merkl/MerklClaimRewardHook.sol
0xTimepunk: I don't think it makes sense to claim reward tokens and lock them in vault bank, so this can be removed

Can confirm with @vikramarun

src/hooks/claim/merkl/MerklClaimRewardHook.sol
0xTimepunk: I think because the function placeholders were added

`    /// @inheritdoc ISuperHookInflowOutflow
    function decodeAmount(bytes memory) external pure returns (uint256) {
        return 0;
    }

    /// @inheritdoc ISuperHookContextAware
    function decodeUsePrevHookAmount(bytes memory) external pure returns (bool) {
        return false;
    }

    /// @inheritdoc ISuperHookOutflow
    function replaceCalldataAmount(bytes memory data, uint256) external pure returns (bytes memory) {
        return data;
    }`
    
    if they are not needed can be removed

src/hooks/claim/merkl/MerklClaimRewardHook.sol
0xTimepunk: I think no

src/hooks/claim/merkl/MerklClaimRewardHook.sol
TamaraRingas: Oh thanks this is from when we still had users as a data input, updating

src/hooks/claim/merkl/MerklClaimRewardHook.sol
TamaraRingas: users array is created by the hook and matches the input array length for all 4 arrays so won't be 0 for just users and if they are all 0 it will revert

0xTimepunk (approved):

PR #737, opened 2025-08-01
fix: EIP 1271 support

0xTimepunk:
## Security Improvements

- Replaced custom ECDSA validation with OpenZeppelin's battle-tested ECDSA library
- Improved error handling for invalid signatures

## Cross-Chain EIP-1271 Support

- Implemented chain-agnostic domain separator with fixed chain ID (1) for cross-chain compatibility
- Added hybrid validation approach that supports both:
  - Chain-agnostic signatures (works across all chains)
  - Native Safe signatures (works with existing Safe UI)
- Created fallback mechanism for seamless Safe UI compatibility

## Implementation Highlights

- Uses OpenZeppelin's ECDSA.tryRecover() for comprehensive security checks
- Maintains signature ordering validation to prevent signature reuse
- Supports both same-chain and cross-chain operations with a single codebase
- Enables progressive UX enhancement without breaking existing workflows

### DIFF
--- a/src/libraries/ChainAgnosticSafeSignatureValidation.sol
+++ b/src/libraries/ChainAgnosticSafeSignatureValidation.sol
@@ -11,6 +11,7 @@ import { ISuperValidator } from "../interfaces/ISuperValidator.sol";
 library ChainAgnosticSafeSignatureValidation {
     /// @notice Chain-agnostic domain separator type hash
     /// @dev Uses a fixed domain without chainId for cross-chain compatibility
+    /// @notice verifyingContract is the safe smart account address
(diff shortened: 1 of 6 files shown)


src/libraries/ChainAgnosticSafeSignatureValidation.sol
```diff
@@ -12,7 +13,7 @@ library ChainAgnosticSafeSignatureValidation {
     /// @dev Uses a fixed domain without chainId for cross-chain compatibility
     // keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)")
```
sujithsomraaj, line 14: this is not the verifying contract, it should be the safe

TamaraRingas (approved):

PR #751, opened 2025-08-11
feat: increase circle branch coverage

TamaraRingas:
## `CircleGatewayMinterHook.sol` Problems fixed
- Adds verification that ensures all attestations in an AttestationSet reference a single token address.
- Adds `destinationCaller` address validation
- Sets `asset` address in `_preExecute()` function and reads from it in `_postExecute()` 

## Coverage 
This PR also increases the branch and function coverage for both of the Circle Gateway hook contracts and adds an integration test with deposits on multiple source chains and a mint on a destination chain via an AttestationSet. 

### Coverage before PR
Screenshot 2025-08-11 at 19 00 41

### Coverage after PR
Screenshot 2025-08-12 at 12 49 57


src/hooks/bridges/circle/CircleGatewayMinterHook.sol
```diff
@@ -189,25 +187,64 @@ contract CircleGatewayMinterHook is BaseHook {
...
+        while (!cursor.done) {
+            attestation = cursor.next();
+
+            // Extract and validate the `TransferSpec`
+            bytes29 spec = attestation.getTransferSpec();
+
+            // Ensure the caller is the specified destination caller
+            address destinationCaller = AddressLib._bytes32ToAddress(spec.getDestinationCaller());
+
+            if (destinationCaller != account && destinationCaller != address(0)) {
```
0xTimepunk, line 245: I believe we should only revert if (destinationCaller != account)

That is because if a destinationCaller is address(0) it won't cause any problems by front-running

src/hooks/bridges/circle/CircleGatewayMinterHook.sol
```diff
@@ -189,25 +187,64 @@ contract CircleGatewayMinterHook is BaseHook {
...
 
-        // Extract the destination usdc from the transfer spec
-        bytes32 usdcBytes32 = TransferSpecLib.getDestinationToken(transferSpec);
+            // Extract and validate the `TransferSpec`
+            bytes29 spec = attestation.getTransferSpec();
+            destinationToken = AddressLib._bytes32ToAddress(spec.getDestinationToken());
 
-        // Convert bytes32 to address
-        usdc = AddressLib._bytes32ToAddress(usdcBytes32);
+            if (token != address(0) && token != destinationToken || destinationToken == address(0)) {
```
0xTimepunk, line 214: This is a rather complex if statement, can it be broken down? Is it adequately tested? 

I assume this is to test that all destinationTokens are equal

I also feel the error should be "DESTINATION_TOKENS_NOT_UNIQUE" or similar

src/hooks/bridges/circle/CircleGatewayMinterHook.sol
```diff
@@ -189,25 +187,64 @@ contract CircleGatewayMinterHook is BaseHook {
         signature = BytesLib.slice(data, offset, signatureLength);
     }
 
-    /// @notice Extract usdc address from the attestation payload
+    /// @notice Extract token address from the attestation payload
     /// @param data The hook data containing attestation payload and signature
-    /// @return usdc The destination usdc address from the attestation
-    function _extractTokenFromAttestation(bytes memory data) internal pure returns (address usdc) {
+    /// @return token The destination token address from the attestation
+    function _extractTokenFromAttestation(bytes memory data) internal pure returns (address token) {
```
0xTimepunk, line 193: Have we introduced a test for "multi-chain USDC burns aggregated into a single mint on the destination chain"

0xTimepunk (changes requested):

PR #754, opened 2025-08-12
fix: withHook bypass scenario

0xTimepunk:
- Fixes withHook bypass scenario by requiring the callee to resetExecutionState to be lastCaller

Note: any code that uses hooks outside of a SuperExecutor context cannot execute that code within a SuperExecutor context


test/integration/E2EExecution.t.sol
```diff
@@ -29,6 +30,55 @@ import { MockMaliciousHook } from "../mocks/MockMaliciousHook.sol";
 import "forge-std/console2.sol";
 import "forge-std/Test.sol";
 
+contract MaliciousHookResetExecution {
```
supervaulter, line 33: Small thing. This should stay in the `mocks` folder

PR #778, opened 2025-08-19
patch: update slippage protection for swappers

supervaulter:


0xTimepunk (commented):
Code is repeated across 3 files, this could easily be a small library that each of the 3 hooks import. Also, add unit tests for it

PR #787, opened 2025-08-22
fix: revert commit 1fb265a and Detect flavors appropriately

ghost:
This reverts commit 1fb265a24b3df9861a659a2e0fc97cb13be521ee.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,16 @@ Handles accounting aspects (pricing, fees) for both INFLOW and OUTFLOW operation
 
 The system uses a dedicated on-chain oracle system to compute the price per share for accounting. Specialized oracles exist for different vault standards (ERC4626, ERC5115, ERC7540, etc.) that provide accurate price data and TVL information.
 
+**⚠️ PPS Unit Semantics & Pendle Gotchas**
+
+Not all `getPricePerShare()` values are returned in underlying asset units:
+
+- **ERC-4626/ERC-7540/Spectra PT**: Returns base-asset denominated amounts (in underlying token decimals)
+- **Pendle PT**: Returns ratio from `IPMarket.getPtToAssetRate()` scaled to 1e18 (assets per 1 PT)
+- **ERC-5115**: Returns `exchangeRate()` scaled to 1e18 (assets per share)
+
+`SuperYieldSourceOracle` normalizes ratio-style PPS (1e18) to base token units only for quote functions via auto-detection. Accounting remains consistent by design. FlatFeeLedger is used for ERC-5115 and Pendle PT.
+
 ### Infrastructure
 
 #### SuperBundler
--- a/foundry.lock
+++ b/foundry.lock
@@ -1,0 +1,44 @@
+{
+  "lib/ExcessivelySafeCall": {
+    "rev": "81cd99ce3e69117d665d7601c330ea03b97acce0"
+  },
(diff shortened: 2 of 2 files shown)


src/accounting/oracles/SuperYieldSourceOracle.sol
```diff
@@ -251,4 +258,56 @@ contract SuperYieldSourceOracle is ISuperYieldSourceOracle {
...
+        }
+    }
+
+    function _detectFlavor(address yieldSourceAddress) internal view returns (Flavor) {
+        // Probe 1: Pendle Market (has readTokens())
+        try IPMarket(yieldSourceAddress).readTokens() {
+            return Flavor.PendlePT;
+        } catch { }
+
+        // Probe 2: ERC-5115 (has exchangeRate())
```
supervaulter, line 281: Is there a possibility for a market to have `readTokens() & exchangeRate()` but to be treated as 5115 instead of Pendle?Or we ACK these cases?

src/accounting/oracles/SuperYieldSourceOracle.sol
```diff
@@ -251,4 +258,56 @@ contract SuperYieldSourceOracle is ISuperYieldSourceOracle {
...
+        returns (uint256 baseAmount)
+    {
+        uint256 pps = IYieldSourceOracle(yieldSourceOracle).getPricePerShare(yieldSourceAddress);
+        Flavor f = _detectFlavor(yieldSourceAddress);
+
+        if (f == Flavor.PendlePT || f == Flavor.ERC5115) {
+            // pps is a ratio (assets/share) scaled 1e18. Convert to base units:
+            // baseAmount = pps * 10^baseDecimals / 1e18
+            uint8 baseDec = _safeBaseDecimals(base);
+            return Math.mulDiv(pps, 10 ** uint256(baseDec), 1e18);
```
supervaulter, line 307: Should we round down here? (not to undercharge). For example, in case of a `repay` in lending markets, it's usually rounded up for this reason. Here might be a case when we need to round it up. Wdyt?

src/accounting/oracles/SuperYieldSourceOracle.sol
```diff
@@ -251,4 +258,56 @@ contract SuperYieldSourceOracle is ISuperYieldSourceOracle {
...
+    function _safeBaseDecimals(address base) internal view returns (uint8) {
+        // Fallback to 18 if token is non-standard or reverts
+        try IERC20Metadata(base).decimals() returns (uint8 d) {
+            return d;
+        } catch {
+            return 18;
+        }
+    }
+
+    function _detectFlavor(address yieldSourceAddress) internal view returns (Flavor) {
```
supervaulter, line 275: This is now pretty generic and I think it's problematic to be at the `SuperYieldSourceOracle` level, which is supposed to work with any kind of oracle.
Any `yieldSourceAddress` implementing `readTokens` or `exchangeRate` will be interpreted as either Pendle or 5115. I can have a random defi vault that has an `exchangeRate` method that can return in 1e8 (similar to chainlink oracles for example) with no relation to `base`'s decimals, even if it's not 5115, it will be interpreted as such

supervaulter:
We discussed and decided to ACK the fact this cannot cover the flavors 100% Approving

supervaulter (approved):

PR #805, opened 2025-09-08
feat: hook proposal changes

ghost:


sujithsomraaj:
So if a smart account sets an operator, it can bypass fees and if we force include these claims then the transaction could fail if the operator front-runs.

Also users can set only operators can claim, DoSing the entire flow / action through superform if this transaction is added to the execution

PR #860, opened 2026-03-09
feat: added PSM integration hooks [SV-1516]

supervaulter:


NicolaBernini:
# Analysis 

One minor thing 

## The `inspect()` returns misleading receiver

All 4 hooks' inspect() returns (assetOut, receiver) where receiver is read from data offset 104. But the actual execution ignores the receiver and forces it to account.
Off-chain systems consuming inspect() output would see a receiver that doesn't match the actual execution behavior.

SwapSparkPSMExactInHook.sol:121-125:

```
  function inspect(bytes calldata data) external pure override returns (bytes memory) {
      address assetOut = data.toAddress(20);
      address receiver = data.toAddress(104);  // ← this is IGNORED in execution
      return abi.encodePacked(assetOut, receiver);
  }
```

Compare with Odos hooks which return (executor) — a different pattern but at least consistent with what the hook actually uses.

Suggestion: Either return account as the receiver (but inspect is pure so can't), or document clearly that the receiver in inspect output is informational only, or just return assetOut without the misleading receiver.

NicolaBernini (dismissed):
Overall LGTM, just one minor thing mentioned here 
https://github.com/superform-xyz/v2-core/pull/860#issuecomment-4025605237

NicolaBernini (approved):
LGTM

PR #862, opened 2026-03-19
feat: kyberswap hooks deployment + inspect update [SV-1535]

supervaulter:
## Summary
- Added KyberSwap hooks (`SwapKyberSwapHook`, `ApproveAndSwapKyberSwapHook`) to deployment scripts
- Updated `inspect()` to return only `dstToken` instead of all 4 addresses
- Generated and locked bytecode for prod and staging
- Deployed to all supported chains (prod + staging)

Replaces #861 (rebased cleanly on latest dev)

## Test plan
- [x] `forge build` passes
- [x] Unit tests updated for new `inspect()` return value
- [ ] Verify deployment addresses in output JSON files


NicolaBernini (approved):
Overall LGTM but would like to call 2 things 

1) Here inspect() has been updated to return only dstToken which is an address so 20 bytes while previously it was returning 3 addresses so 80 bytes 
If there is a consumer that is expecting 80 bytes, this is a breaking change 

2) Looks like the 3 aggregator hooks all return different things since 
Kyber returns 1 address which is dstToken 
Odos returns 1 address but it is executor 
and 1Inch returns various addresses 

Again, this is not an issue in itself as long as the consumer is aware of this lack of consistency between hooks belonging to the same category i.e. aggregators

PR #872, opened 2026-04-09
chore: fixes for morpho hooks [SUP-19387]

supervaulter:


NicolaBernini:
LGTM Overall but we need to also add the Flashloan Hook 

The reason is while it is technically possible to build a leverage borrow position by doing N times (deposit collateral, borrow, swap, repeat) nobody in reality does that since it is highly gas inefficient: the standard approach is the Contango one which involves using a flashloan and allows to build a final position which is totally equivalent to the one built in the naive way with a single swap(), saving a lot of gas in the process 

So there are 2 workstreams to support this: onchain and offchain 

1) Onchain 
On Flare we have Morpho, so the hook to add is `Morpho.flashLoan(token, assets, data)` and since it is a flash loan this involves handling also the callback `onMorphoFlashLoan(assets, data)` 

2) Offchain 
Superman needs to use this hook to build the leveraged position

PR #892, opened 2026-05-12
feat: added DETH hooks [SUP-19747]

supervaulter:
E2E tests in test/integration/deth/DETHOracleAndHooksE2E.t.sol:                                                                                                                                                                          
                                                                                                                                                                                                                                            
  Oracle sanity (8):                                                                                                                                                                                                                          
  - decimals, PPS matches Machine, TVL matches Machine, strategy balance, held-only TVL, zero address, round-trip, conversion consistency                                                                                                     
                                                                                                                                                                                                                                              
  Pending TVL tracking (3):                                                                                                                                                                                                                   
  - pending after requestRedeem, multiple requests accumulate, filtered by owner                                                                                                                                                            

  Full lifecycle (4):
  - hold → request → pending TVL → claim → final TVL
  - real claim on finalized request with WETH delta tracking
  - requestRedeemHook variant
  - strategy rebalance with 2 sequential partial redeems

  FOUNDATION routing (12):
  - routed redeem via safe address (full PPS-preserving flow)
  - SuperWETH PPS stability through routed redeem
  - without foundation → TVL drops
  - only counts router pending, not other addresses
  - mixed pending (strategy + router both counted)
  - multiple routed requests accumulate
  - router own TVL no double-count
  - full redeem all pending via router
  - immutable is set, zero address reverts
  - PPS comparison with vs without foundation
  - PPS stable through incremental routed redeems

  Withdrawal rounding (1):
  - ceil rounding verification


supervaulter:
@claude review

src/accounting/oracles/DETHYieldSourceOracle.sol
```diff
@@ -0,0 +1,314 @@
...
+        uint256 totalPendingShares;
+
+        for (uint256 i; i < scanLimit; ++i) {
+            uint256 requestId = lastFinalized + 1 + i;
+
+            // Check if this request belongs to either owner
+            try IERC721(asyncRedeemer).ownerOf(requestId) returns (address nftOwner) {
+                if (nftOwner == primaryOwner || (checkSecondary && nftOwner == secondaryOwner)) {
+                    // Get shares locked in this request
+                    try redeemer.getShares(requestId) returns (uint256 shares) {
```
NicolaBernini, line 294: This is not a bug but worth a clarification 
The redeemed shares are converted into assets at request time 
https://github.com/MakinaHQ/makina-periphery/blob/main/src/redeemers/AsyncRedeemer.sol#L138

so, at a first sight, this implementation looks inaccurate 

However, in practice, this onchain calculated amount of assets is not really binding for them since the amount of assets used to actually fulfill is recalculated when the pending request is redeemed 
https://github.com/MakinaHQ/makina-periphery/blob/main/src/redeemers/AsyncRedeemer.sol#L183

and it is based on the current PPS
https://github.com/MakinaHQ/makina-core/blob/main/src/machine/Machine.sol#L470

So the current implementation is actually accurate

src/accounting/oracles/DETHYieldSourceOracle.sol
```diff
@@ -0,0 +1,314 @@
...
+        } catch {
+            return 0;
+        }
+
+        // No pending requests
+        if (nextId <= lastFinalized + 1) return 0;
+
+        bool checkSecondary = secondaryOwner != address(0);
+        uint256 pendingCount = nextId - lastFinalized - 1;
+        uint256 scanLimit = pendingCount > MAX_PENDING_REQUESTS ? MAX_PENDING_REQUESTS : pendingCount;
```
NicolaBernini, line 284: This can be gas expensive but if this function is only called from offchain logic i.e. the SuperVault PPS calculation function, then it should be OK

NicolaBernini (approved):
LGTM 
Just a couple of comments

PR #894, opened 2026-05-15
feat: added rFLR claim hook [SUP-19774]

supervaulter:


supervaulter:
@claude review

NicolaBernini:
# Overview

Good Stuff
- Pattern-faithful: BaseHook inheritance, HookType.NONACCOUNTING + HookSubTypes.CLAIM, balance-delta with safe zero-clamp on decrease, immutable targets — same shape as MerklClaimRewardHook.
- MAX_PROJECT_IDS = 50 caps gas-griefing surface; immutable constructor validation rejects zero addresses.
- Test coverage is great for a small hook: unit + fork-integration on real Flare mainnet holders with real claimable balances + invariant suite (7 invariants, ghost state, mint/burn handlers that keep totalSupply consistent) + a SuperVault E2E. This is more rigour than I usually see from this author and worth a positive note.
- Security report is included in-tree (good practice).
- Deployment scripts gate the rFLR hooks to FLARE_CHAIN_ID = 14 so they cannot be mis-deployed.

About the concerns, see the following analysis

# Analysis

## High Priority 

### M-2 Slippage Check 

Regarding the last Claude bot review here 
https://github.com/superform-xyz/v2-core/pull/894#issuecomment-4459824127

The M-2 is imho the more concerning thing due to my view about slippage checks 

Let me start more at a high and "philosophical" level 

My historical positioning with slippage protection is that slippage checks 
- must be onchain enforced, the offchain check is a nice to have, meaning it's good to have it but this is not a replacement for the onchain check since it does not provide the same level of guarantees that an onchain check provides (in fact, it is absolutely possible that slippage does not manifest itself in the offchain check but only in the onchain one e.g. due to MEV)
- are required everywhere it is possible to have them 

Now let's apply this to the specific case 

First of all, we need to observe here we should have actually 2 hooks in scope 
H1: ClaimRFLRHook
H2: WithdrawRFLRHook

The H1 is a claim-type hook and therefore by construction it should have no slippage source it (meaning if user is eligible to claim X, then he can claim X, no MEV or other tricky paths possible)

The H2 also can't have "slippage" in its strict definition, but still it can have a loss so the same principle of slippage should be applied
The whole point of slippage protection enforced onchain is to have an immutable evidence of the user consenting to accept up to a certain X% loss, regardless of it coming from slippage in a classic sense or other economic mechanisms 
So even though WithdrawRFLRHook does not have slippage in a classic sense, it can still originate losses and therefore we need to offer a mechanism where an onchain consent to accept this loss is provided 

Two proposals to consider 

Variant A — minOut on WFLR delta (swap-hook style):
```
function _postExecute(address, address account, bytes calldata data) internal override {
      uint256 delta = IERC20(WFLR).balanceOf(account) - getOutAmount(account);
      if (data.length >= 32) {
          uint256 minOut = BytesLib.toUint256(data, 0);
          if (delta < minOut) revert SLIPPAGE_EXCEEDED();
      }
      _setOutAmount(delta, account);
  }

```


  Variant B — hard guard against locked burn (pre-execute):
```
function _preExecute(address, address account, bytes calldata data) internal override {
      asset = WFLR;
      (,, uint256 locked) = IRNat(RNAT).getBalancesOf(account);
      if (locked > 0) {
          bool ack = data.length >= 1 && data[0] != 0;
          if (!ack) revert UNVESTED_BURN_NOT_ACKNOWLEDGED();
      }
      _setOutAmount(IERC20(WFLR).balanceOf(account), account);
  }  
```
  
  
  (A) matches the swap-hook convention; the caller specifies the exact WFLR floor they'll accept. (B) is the strictest — caller must explicitly opt-in to any locked burn. 
  They're not mutually exclusive; you could ship both, with data[0] as the acknowledge bit and data[1:33] as the optional minOut. One extra SLOAD on Variant B, one comparison on Variant
  A. Negligible gas.


My preference would be to have both and by default keep A active only and make B a no op initially but we can enable it in case some curators need tighter protection 



### H-1 destination validAfter ignored → premature WithdrawRFLRHook → irreversible burn

I can't see the Octane detailed report since I am not authorized to access the app, however, it looks like something pre-existing and not anything this PR has introduced so I'd rather not block this PR on this finding and solve it in a different PR 




## Mid 

### The `WithdrawRFLRHook` silently ignores all input data

Looks like 

```
function _buildHookExecutions(address, address, bytes calldata) internal view override
```

The bytes calldata parameter is fully unused.

Cheap hardening: 

```
if (data.length != 0) revert UNUSED_DATA();
```


### M-3 Cancun Compatibility 

I think Claude is right flagging this and it's good to have a strong reference supporting it 
After some research, I have found this 

> Supported opcodes: Supports all EVM [opcodes](https://www.evm.codes/?fork=cancun) up to and including the Cancun hard fork.

https://dev.flare.network/network/overview

which should be a strong enough evidence that Cancun is supported 



### M-1 

Seems an easy fix 


### The `IRNat.claimRewards` seems wrong 

It says 

```
/// @return claimedAmount Total WFLR deposited into caller's RNat account
function claimRewards(uint256[] calldata projectIds, uint256 month) external returns (uint128 claimedAmount);
```

but `claimRewards` should mint rFLR (the non-transferable wrapper), not WFLR.

This is misleading for AI and integrators 



## Minor

### H-2 NONACCOUNTING + chainable → fee bypass

This does not look like a real issue since fees are charged on the outflow boundary and not on rewards landing

NicolaBernini (changes requested):
Overlla looks good but I recommend taking a look at my M-2 comment here 
https://github.com/superform-xyz/v2-core/pull/894#issuecomment-4485978760

supervaulter:
In terms of `The WithdrawRFLRHook silently ignores all input data` -> the data is used in pre and post execute so it works correctly. It's just not used in buildExecution, but the data is still needed

Cancun compatibility is ok yes.

Correcting natpsec for claimRewards and adding the slippage protection you mentioned

NicolaBernini (approved):
OK to proceed

PR #899, opened 2026-05-22
feat: added deallocate hook [SUP-19870]

supervaulter:


NicolaBernini:
# Overview 

Since the `forceDeallocate` does not pull any token from the user no approval should be needed 

Try to remove it in the tests and see if it still works

supervaulter:
> # Overview
> Since the `forceDeallocate` does not pull any token from the user no approval should be needed
> 
> Try to remove it in the tests and see if it still works

Good point. Removed

PR #907, opened 2026-05-29
feat: metavault oracle [SUP-19953]

supervaulter:


NicolaBernini:
# Overview 

Overall LGTM but there are a few non-blocking fixes

# Details

## Core is solved 

The key issues have been addressed, let me recap them 
- getTVL (oracle L182-186): vault.convertToAssets(IERC20(shareToken).totalSupply()). Avoids OZ's default totalAssets() (idle balance).
- in the `getTVLByOwnerOfShares()` the Component 3 / claimable redeem (oracle L243-249): claimableRedeemRequest(REQUEST_ID, owner) → convertToAssets(claimableShares), not maxWithdraw. This is the core fix.
- Hardening vs the generic ERC7540YieldSourceOracle: inner convertToAssets calls are now wrapped in try/catch (L234, L245); getWithdrawalShareOutput reverts on PPS=0 (L101, require); _getShareToken guards share()==address(0) (L272-278). Components 1/2/4/5, PPS, and share-token discovery are unchanged.


## Static Proof of no double counting

The 5-component sum is correct iff held / pending-redeem / claimable-redeem are mutually exclusive. Verified from the deployed Flare MetaVaultWrapper (impl 0x2151):

Held ∩ {pending, claimable} = ∅ — requestRedeem burns the shares (_burn(owner, shares), MetaVaultWrapper L566), so redeemed shares leave balanceOf (Component 1) and move into the redeem-request accounting. (This is the bug the old maxWithdraw path had — it re-read balanceOf and double-counted held; now eliminated.)

Pending ∩ claimable = ∅ — pendingRedeemRequest reads the current infra epoch c (L308-316); claimableRedeemRequest reads the last-request epoch r and returns 0 when r == c (L319-325 → _amphorPreviewClaimRedeem L727-744). And requestRedeem auto-claims any prior claimable before booking a new request (L549-552), setting r = c. So a non-zero pending (epoch c) forces r == c → claimable = 0, and vice-versa.

State-by-state (owner O = SuperVault strategy account; R = last-settled rate; B = held shares; S = redeemed shares):

State | C1 held | C2 pending | C3 claimable | getTVLByOwnerOfShares | TRUE
-- | -- | -- | -- | -- | --
Steady hold | B·R | 0 | 0 (r=0) | B·R ✓ | B·R
Redeem S, pending (c=e0) | (B−S)·R | S·R | 0 (r==c) | B·R ✓ | B·R
Redeem S, settled/claimable (c=e0+1) | (B−S)·R | 0 | S·R | B·R ✓ | B·R
Full redeem (S=B), claimable | 0 | 0 | B·R | B·R ✓ | B·R


## The non-blocking fixes recommended

### [F1 — MEDIUM, verify + doc] The PR was developed against the Base USDC MetaVault; 
bizFXRP will use the Flare FXRP instance — same address, different behavior. The tech spec, security report, and oracle NatSpec all state "MetaVaultWrapper does not override totalAssets() → idle USDC (0)" (techspec L9-19, "Verified On-Chain Facts" = Base: convertToAssets(1e6)=863399, totalAssets()=0). But the deployed Flare 0x6420 (FXRP, impl 0x2151) does override totalAssets() → convertToAssets(totalSupply()) (L358), so totalAssets() is non-zero there — the Flare fork test's own header says so (L16-19), and test_fork_flare_getTVL_genericOracle_nonZero confirms even the generic oracle's getTVL is non-zero on Flare. The fix is correct on both (convertToAssets(totalSupply) == totalAssets on Flare; claimableRedeemRequest is the right source regardless of the totalAssets/maxWithdraw values). But the same oracle now serves two behaviorally-divergent deployments. Action: (a) fix the NatSpec/spec to say "Base does not override totalAssets; Flare does; convertToAssets(totalSupply) is correct for both"; (b) confirm the two MetaVaultWrapper impls are the same version family beyond the two functions checked (esp. claimableRedeemRequest units = shares, and convertToAssets epoch semantics), so the oracle can't silently misprice one.


### [F2 — MEDIUM, the main substantive gap] No test exercises a real redeem lifecycle on either chain. 
Both fork tests (SpectraMetaVaultOracleFlare.t.sol, …Fork.t.sol Base) are view-only on empty/zero/0xdead positions — no requestRedeem/epoch-advance/settle/prank/deal. The pending & claimable redeem components are populated only in unit tests via MockSpectraMetaVault.setPendingRedeemShares/setClaimableRedeemShares, where the author can set combinations the real contract forbids (e.g. balanceOf>0 AND claimableRedeemRequest>0). So test_allFiveComponents proves the summation arithmetic, not that the real contract keeps components disjoint — which is precisely the security report's own P2-3 ("needs on-chain verification"), still open empirically. The §2 proof closes it statically, but a fork test should demonstrate it: impersonate a gamisXRP holder, requestRedeem, advance the infra epoch + settle, assert the breakdown is disjoint and sum == B·R at each step, then redeem (claim).

### [F3 — LOW, the P2-2 epoch-rate nuance, now precise] Claimable is priced at the last-settled epoch, not the fulfillment epoch. 
Component 3 uses 1-arg convertToAssets(claimableShares), which prices at epoch c−1 (MetaVaultWrapper L221-230), whereas the actual claim realizes at the request's fulfillment epoch r (_claimPendingRedeem uses convertToAssets(_, r), L861). For prompt claims (r == c−1) this is exact; otherwise it over-/under-values by the inter-epoch yield drift (~0.014%/epoch at ~4.5% APY, ~28h epochs) — immaterial in practice, overvaluation direction. The security report rates this P2-2 "likely mitigated, needs verification"; the precise read is "real but bounded." Exact fix (optional): read lastUserRedeemRequestEpoch(owner) = r and use the 2-arg convertToAssets(claimableShares, r) (MetaVaultWrapper L233).

### [F4 — LOW, process] The internal security report (2026-05-28) is stale vs the shipped code. 
It lists as open/accepted several items already implemented in the PR head: P2-1 (inner convertToAssets try/catch — done, L234/245), P3-2 (getWithdrawalShareOutput revert on PPS=0 — done, L101), P3-3 (share()==address(0) guard — done, L272-278), P3-5 (catch-block comments — done). Its summary (3×P2 / 5×P3) overstates open issues. Net positive (code is better than the report claims), but regenerate the scan against the PR head so reviewers aren't misled, and so P2-3 is reconciled with the §2 proof.

### [F5 — LOW, doc] getTVL ≠ Σ getTVLByOwnerOfShares in flight. getTVL = convertToAssets(totalSupply) excludes in-flight redeems (shares burned, not in totalSupply) and pending/claimable deposits (assets not yet minted). 
The oracle comment + report P3-1 mention only deposits — add redeems. Low severity (getTVL is not on the pricing path: supervault-pricing/app/services/yield_source.py uses getTVLByOwnerOfShares[Multiple]), but ensure no monitoring invariant assumes getTVL == Σ per-owner.

### [F6 — LOW, availability] Inner convertToAssets try/catch → silent 0 on revert. 
Good that it no longer DoS-reverts the whole breakdown, but a convertToAssets revert now silently drops the pending/claimable redeem component to 0 → undercount (PPS-dip direction). Practically unlikely (epoch c−1 is always settled for c ≥ 2), but the off-chain circuit breaker should treat a sudden single-sleeve TVL drop as "pause/await settle," not publish.

### [F7 — INFO, scope] Unrelated binding churn. 
The PR bundles contract_bindings/NativeFeeSponsorship/… and SuperNativePaymaster regeneration (from the modified generate-contract-bindings.sh). Confirm intentional/harmless; ideally isolate to reduce review noise.

### [F8 — INFO, follow-ups not in this PR] Deploy + wire. 
The PR adds the contract + locked bytecode + bindings + tests, but no deployment or registration (no script/output/prod/14 change, no SuperLedgerConfiguration entry). To onboard gamisXRP: (1) deploy SpectraMetaVaultOracle on Flare (and Base if those MetaVaults are onboarded); (2) register vault→oracle→ledger in SuperLedgerConfiguration (propose → timelock → accept); (3) point the off-chain supervault-pricing service at this oracle for gamisXRP (not erc7540, not spectra_pt).

NicolaBernini (approved):
Overall LGTM with a few non-blocking recommendations 

See the details here 

https://github.com/superform-xyz/v2-core/pull/907#issuecomment-4574761793

PR #908, opened 2026-05-29
feat: openocean sparkdex hook

subhasishgoswami:


supervaulter:
Some small comments

`_updateMakeCall` calls `_updateCallData` and we don't have any enforcement on the depth limit. Just wanted to flag; since the bundler control the depth shouldn't be an issue, but for protection we could also enforce a max iteration there

There's a mismatch for `value` and `inputAmount`. When usePrevHook is false (and as far as I understood from the chat in slack this will always be the case), `value` is used directly from hook data without a validation against `inputAmount`. So excess ETH can be lost. We can derive the value similar to Odos V3 hook `inputToken == address(0) ? inputAmount : 0`

For ` ApproveAndSwapOpenOceanSparkDexHook`, should we support native path ? It currently creates 4 execution paths no matter what. In Odos v3, I'm doing `if (inputToken == address(0))` and have a separate branch for this case


Nits:

Validation for `inputToken == outputToken` and revert (like Odos v3)
`CallDescription.target ` not validated, should we?
Missed `ISuperHookInspector` import

supervaulter (approved):

PR #916, opened 2026-06-02
feat: stargate adapter updates

supervaulter:
## Problem                                                                                                                                                                    
                                                                                                                                                                                 
  The previous `StargateAdapter.lzCompose` implementation had several issues:                                                                                                    
                                                                                                                                                                                 
  1. **Reverting lzCompose blocks the pipeline** — LayerZero compose messages are ordered per source. If `lzCompose` reverts (e.g. from `processBridgedExecution` or a failed    
  ERC20 transfer), ALL subsequent composes from that source are permanently blocked.                                                                                             
                                                            
  2. **No recovery for failed transfers** — If the token transfer to the account failed, the transaction reverted and tokens stayed at the adapter with no claim mechanism.

  3. **Zero-address account = permanent fund loss** — If `account == address(0)` was encoded in the compose payload, there was no handling — tokens were lost.

  4. **Token resolution failure = full revert** — If `_from.token()` reverted (invalid Stargate/OFT contract), the entire compose reverted and blocked the pipeline.

  ### Changes

  #### `src/adapters/StargateAdapter.sol`

  - **Non-reverting design** — `lzCompose` no longer reverts after sender validation. All failure paths are handled gracefully to avoid blocking the LZ compose pipeline.

  - **try/catch on `_from.token()`** — Token resolution is wrapped in try/catch. On failure, emits `TokenResolutionFailed` and returns early.

  - **try/catch on `processBridgedExecution`** — Execution failures emit `ExecutionFailed` instead of reverting. Execution is always attempted (even if transfer failed) to
  consume the signed merkle root and prevent stale intents.

  - **`_tryTransfer` helper** — Replaces direct `safeTransfer`/ETH transfer with a non-reverting transfer that returns success/failure. Uses low-level call for ERC20s to support
   non-standard tokens (USDT) that don't return bool.

  - **`failedTransfers` mapping + `claimFailedTransfer()`** — When token transfer fails, the amount is stored in `failedTransfers[account][token]`. The account can later call
  `claimFailedTransfer(token, amount)` to recover. Protected by `ReentrancyGuard`.

  - **`composeFrom` fallback for zero-address account** — Extracts `composeFrom` from OFTComposeMsgCodec header (bytes 44-76, the source chain sender). When `account ==
  address(0)`, stores tokens in `failedTransfers[composeFrom][token]` so the original sender can claim on the destination chain.

  - **New events** — `TransferSucceeded`, `TransferFailed`, `ExecutionFailed`, `TokenResolutionFailed`, `FailedTransferClaimed` (replaces the old `ComposeExecuted`)

  #### `test/integration/stargate/StargateAdapterE2EFork.t.sol` (new)

  Part of the E2E fork tests bridging ETH→Base through real Stargate infrastructure with Pigeon for LZ message relay:

  1. **Happy path** — full bridge + destination execution + root marked
  2. **Execution failure** — wrong signature, tokens still safe at account
  3. **Transfer failure** — tokens stored in `failedTransfers`, recovered via `claimFailedTransfer()`
  4. **Zero-address account** — tokens stored under `composeFrom`, sender claims


supervaulter:
@claude review this please

subhasishgoswami (approved):

PR #923, opened 2026-06-12
feat: hooks standardization

supervaulter:
## Summary

Hook standardization pass: introduces the sizing interface, pipe-mode architecture, manifest tooling, and CI quality gates across all hooks. This is an audit-relevant change — it touches BaseHook, ISuperHook, all ~120 hook contracts, and the off-chain manifest pipeline.

---

### Interface layer (`src/interfaces/ISuperHook.sol`)

- **`ISuperHookInflowOutflow`** — new: `decodeAmounts(bytes)` (array), `amountRoles(bytes)` (array of `AmountMeta{Direction, Denomination}`)
- **`ISuperHookOutflow`** — new: `replaceCalldataAmounts(bytes, uint256[])` (array form; replaces single-amount predecessor)
- **`ISuperHookContextAware`** — new: `decodeUsePrevHookAmount(bytes)` extracted from base
- **`ISuperHookResult`** — extended: `getOutToken(address)` added alongside existing `getOutAmount`

### BaseHook infrastructure (`src/hooks/BaseHook.sol`)

- **`outToken` transient storage** — per-account ephemeral output token cleared each tx; set via `_setOutToken`, read via `getOutToken`
- **Pipe modes** — `PipeMode.TRANSFORM` (default, balance-delta), `PipeMode.PASSTHROUGH` (no token flow, forwards prev hook's amount/token), `PipeMode.SOURCE` (self-determined amount, ignores prev hook)
- **`_supportsSizingInterface()`** — virtual gate; overriding to `true` declares ISuperHookInflowOutflow/ISuperHookOutflow support
- **Hook metadata** — `name()` and `description()` pure functions added to all hooks (auditor-visible human labels)

### Hook changes (all ~120 hooks)

- All hooks gain `name()` and `description()` implementations
- Sizing hooks implement `decodeAmounts()`, `amountRoles()`, `replaceCalldataAmounts()` in array form
- Loan hooks (Morpho × 7, Aave V4 × 6): `_setOutToken` bug fixed — was setting `address(0)` due to uninitialized transient `asset`; now uses `getLoanTokenAddress(data)` / `getCollateralTokenAddress(data)` per hook semantics
- `MorphoWithdrawHook`: fixed `description()` (was "Withdraws collateral" — wrong; calls `withdraw`, not `withdrawCollateral`), `HookSubTypes.LOAN_REPAY` → `LOAN`, override `decodeUsePrevHookAmount() = false` (offset 144 = `shares` uint256, not a bool)
- `RedeemFirelightVaultHook`: `_setOutToken` fixed to `data.extractYieldSource()` (no asset transfer in this step)

### New hooks

- `ClaimRFLRV2Hook` — safe balance-snapshot variant of ClaimRFLRHook
- `ClaimRFLRV3Hook` — refactored claim with updated encoding
- `WithdrawRFLRHookV2` — Flare rFLR withdrawal V2
- `WithdrawVestedRFLRHookV2` — vested rFLR withdrawal V2

### Manifest tooling (`tooling/`)

| File | Purpose |
|---|---|
| `hook-classification.yaml` | Hand-curated: actionTypes + legSizing per hook |
| `hook-enrichment.yaml` | Off-chain enrichment: protocols, approve pairs, sizelessHooks |
| `generate_hook_manifest.py` | Generates `manifests/hooks.json` from YAML + sol source |
| `lint_hook_manifest.py` | Validates P2 (intent/stage → hookType) + P3 (INFLOW/OUTFLOW → sizing interface) |
| `generate-hook-sizing-manifest.ts` | Generates `hook-sizing-manifest.json` (mode + amountPosition per hook) |
| `validate-hook-sizing-manifest.ts` | Validates sizing manifest completeness + cross-checks offsets against source |

### CI (`lint-manifest` job in `.github/workflows/contracts-ci.yml`)

- **Drift gates**: `git diff --exit-code manifests/hooks.json` after Python generator; `git diff --exit-code hook-sizing-manifest.json` after TS generator — stale committed manifests now fail CI
- **Sizing validator** added to CI (was missing)
- **P3 invariant**: `lint_hook_manifest.py` now asserts `INFLOW/OUTFLOW ⟹ ISuperHookInflowOutflow in erc165` — auto-catches new hooks that forget to wire sizing

### Denomination cleanup

Removed `denom` from `hook-sizing-manifest.json` / the TS generator. Denomination now lives exclusively in `manifests/hooks.json` (derived from on-chain `amountRoles()`). The TS generator's `inferDenom()` heuristic incorrectly classified swap/bridge hooks as `"assets"` — eliminated.

### Repo hygiene

- `.claude/` added to `.gitignore` (AI working files, session logs, internal reviews — must not be committed)
- Removes all previously committed `.claude/` files from the repo tree

---

## Off-chain consumer impact (P2-5 — coordination required)

This PR changes the ABI shape of `ISuperHookInflowOutflow` and `ISuperHookOutflow` from singular to array form. Off-chain consumers **must** be updated in lockstep with deployment:

| Consumer | Impact | Status |
|---|---|---|
| **Superman OMS** | ~53 ABI JSONs stale (ISuperHookInflowOutflow, ISuperHookOutflow, per-hook ABIs); 19 hooks with `replaceCalldataAmounts` need array-aware calldata encoding | ⚠️ pending |
| **SuperBundler** | Highest-risk: calldata encoder must become `amountRoles()`-aware (handle `Direction.IN/OUT`, `Denomination`, XOR slot for MorphoWithdrawHook) | ⚠️ unverified |

**Do not deploy this PR's contracts until off-chain consumers are updated.**

---

## Test plan

- [ ] `forge build` — compiles clean (0 errors, 0 warnings)
- [ ] `make ftest` — all existing tests pass
- [ ] `python3 tooling/generate_hook_manifest.py && python3 tooling/lint_hook_manifest.py` — P2 + P3 checks pass, 124 hooks validated
- [ ] `npx tsx tooling/generate-hook-sizing-manifest.ts` — 120 entries, 0 warnings
- [ ] `npx tsx tooling/validate-hook-sizing-manifest.ts` — 0 errors, 3 expected warnings
- [ ] `git diff --exit-code manifests/hooks.json hook-sizing-manifest.json` — no drift
- [ ] Verify `MorphoWithdrawHook`: constructor uses `HookSubTypes.LOAN`, description correct, `decodeUsePrevHookAmount()` returns `false`
- [ ] Verify loan hook `getOutToken()` returns non-zero after execution for all 14 fixed hooks
- [ ] Off-chain: Superman OMS ABI regeneration + SuperBundler array-shape update verified before merge to main


NicolaBernini:
# Overview

Overall looks like a good architecture but on top of the CI not passing there are also a few findings 


# Findings 

## P1 - Blocking Merge 

### P1-1 · CI lint-manifest RED — 8 classification completeness errors. 

New hooks exist in source but are missing from hook-classification.yaml: ClaimRFLRV2Hook, ClaimRFLRV3Hook, WithdrawRFLRHookV2, WithdrawVestedRFLRHookV2, SwapOpenOceanHook, ApproveAndSwapOpenOceanHook; and 2 stale entries reference deleted source: SwapOpenOceanSparkDexHook, ApproveAndSwapOpenOceanSparkDexHook. Manifest is incomplete; not mergeable until classification is reconciled with source.



### P1-2 · CI coverage RED. 

The new surface (array decode/replace, amountRoles, getOutToken/PipeMode, ERC-165) is under-covered vs the coverage gate. Targeted tests are good but breadth is insufficient — and the gaps are exactly where the silent bugs below hide.


### P1-3 · Two parallel manifests that contradict the on-chain source of truth. 

The hook-sizing-manifest.json (TS, the only one with the 4-mode taxonomy + reason/track/denom) is orphaned — not run in CI. manifests/hooks.json (Python, in CI) carries a different axis (legSizing/amountMeta/addresses/lifecycle). They disagree on denomination, and the TS one disagrees with the contract: e.g. SwapUniswapV3Hook is denom:"assets" in the sizing manifest, but on-chain amountRoles returns TOKEN and manifests/hooks.json says TOKEN — the hand-file is wrong for the entire swap/bridge family. Now that denomination is authoritative on-chain, a hand-curated manifest that contradicts it is the exact "registry #3 / encoding-leak" failure mode we set out to kill. Pick one manifest, derive it from amountRoles.



## P2 - Important 

### P2-1 · getOutToken returns address(0) for ~13 hooks (this looks like a real bug). 

All Morpho + Aave loan hooks and RedeemFirelightVaultHook call _setOutToken(asset, account) in _postExecute, but the transient asset (BaseHook.sol, address public transient asset — defaults to 0 each tx) is never assigned in src/hooks/loan/** or the firelight redeem hook (vault/claim/stake hooks do assign it, e.g. Redeem4626VaultHook.sol:104). Verified: 10+ _setOutToken(asset,…) call sites on pr923, zero asset = assignments in loan/. Doesn't revert, but the new outToken plumbing is silently wrong — a PASSTHROUGH hook chained after a loan/firelight hook forwards address(0). Fix: assign asset, or pass getLoanTokenAddress(data)/getCollateralTokenAddress(data) directly.


### P2-2 · MorphoWithdraw metadata looks like wrong. 

The description() = "Withdraws collateral from a Morpho market", but the hook calls IMorphoBase.withdraw (supplied loan assets, not collateral; collateral withdrawal is withdrawCollateral, used only by RepayAndWithdraw) — _postExecute measuring getLoanTokenBalance confirms the output is the loan token. Live example of the on-chain-immutability cost (P3-1): a copy-paste typo now needs a redeploy to fix. (Pre-existing nits ride along: subtype LOAN_REPAY on a supply-side withdraw; decodeUsePrevHookAmount reads offset 144 = the shares field, a latent collision.)



### P2-3 · Manifest is hand-curated, not generated; no drift gate (R1/R5 not met). 

The sizing data (mode, amountPosition, secondaryAmountPosition, denom, track) is hand-typed in a ~120-entry OVERRIDES dict baked into generate-hook-sizing-manifest.ts; auto-detection is fallback-only. The rigorous validator (validate-hook-sizing-manifest.ts, Check 7 source-offset cross-check) skips inlined offsets (if(sourcePos===null) continue) and isn't wired to CI. The CI job regenerates manifests/hooks.json in place then lints it — no git diff --exit-code, and the "✅ Passed" summary is hardcoded — so a stale committed manifest passes silently. Net: the manifest can drift from contracts undetected, which is precisely the maintenance-tax we rejected.



### P2-4 · No all-hooks conformance invariant (SUP-19991 only partially met). 

Tests assert the interface per enumerated hook instance, but nothing iterates the whole hook set to assert "INFLOW/OUTFLOW ⟹ sizing interface, else NONACCOUNTING" in HookType terms (NONACCOUNTING appears 0× in the test files). A newly added hook won't be auto-caught. Add the universal Foundry invariant (or extend the Python lint to assert it from hookType()).



### P2-5 · Breaking rename = coordinated multi-repo change; array reshape expands M2/M3. 

The v2-core pr923 is internally complete (all 102 in-repo singular sites migrated, 0 left, 1,444 plural occurrences — compiles) and no on-chain non-hook consumer (SuperExecutor, libraries, periphery production) depends on the old names. But off-chain breaks: the Superman OMS ships ~53 stale OLD-interface ABI JSONs (ISuperHookOutflow/ISuperHookInflowOutflow + per-hook ABIs under supervault-pricing/.../abis*, 19 with replaceCalldataAmount) that must be regenerated to the plural/array shape in lockstep with the deploy; SuperBundler (the calldata encoder; local clone empty) is the highest-risk unverified consumer. Crucially, M3's original plan ("call replaceCalldataAmount") must become array + amountRoles-aware (handle Direction.IN/OUT, Denomination, the XOR slot) — more work than the singular plan assumed. Periphery: only a test mock (MockNativeETHHook) needs updating.



### P2-6 · PR hygiene — partner-confidential material committed. 

The diff includes 2,510 lines of .claude/ working files: context_session_17..21.md (5 session logs), agents/hooks-master.md, a CLAUDE.md tweak, and — most seriously — .claude/doc/lotus-hooks-review/findings-report.md (662 lines): a security review of an external partner protocol ("Lotus") with local absolute paths (/Users/cosming/1.Coding/Lotus/src/hooks/loan/lotus/) leaking an unreleased partner's internals (its own CRITICAL/HIGH findings about Lotus inspect() and composite-hook offsets). Strip the entire .claude/** tree from the PR. Also: empty PR description on a 32k-line, 334-file, audit-relevant change.



## P3 — minor / design

### P3-1 · On-chain name()/description() as pure string literals on every hook = bytecode bloat ×120 + immutable (typos → redeploy, cf. P2-2) + duplicated with manifests/hooks.json. 

Reconsider whether UI metadata belongs on-chain; if kept, make the manifest derive from it (single source).



### P3-2 · Two conventions for "authoritatively sizeless." 

Aggregators: supportsInterface(ISuperHookOutflow)==false, no replaceCalldataAmounts. Claims (Fluid/Gearbox/Merkl/Yearn) + permit2 BatchTransferFromHook: implement replaceCalldataAmounts and return length-0 arrays. Both valid (S2), but pick one for reviewability.



### P3-3 · external vs sizeless distinction survives only off-chain. 

On-chain both collapse to "no Outflow interface." The off-chain sizing manifest keeps external (amount in opaque txData; OMS must re-quote, not splice) with a reason — good — but in the orphaned manifest. When consolidating, keep external: the OMS needs "amount exists, can't splice" ≠ "no amount" for telemetry/validation/partial-fill.



### P3-4 · NATIVE_TRANSFER_HOOK_KEY doesn't exist (in fact src/libraries/ defines no *_HOOK_KEY constants). 

A manifest keyed by hookKey can't reference NativeTransferHook — register the key or exclude it explicitly.



### P3-5 · Forward-compat for Midnight: Denomination enum lacks UNITS (Midnight buyWithUnitsTarget vs buyWithAssetsTarget). 

Adding it later is a non-breaking enum extension, but flag now so amountRoles consumers are written tolerant; or model Midnight as per-denomination hooks (Spark-PSM precedent).



### P3-6 · Dead/broken tooling 

The tooling/enrich_manifest.py (368 lines) redundant after enrichment folded into the Python generator; package.json generate:hook-sizing/validate:hook-sizing point at scripts/… but the files are in tooling/ (no scripts/ dir) → both npm scripts fail.


# Recommendations (prioritized)

## R1 - Green the CI before review continues 

Add the 6 missing hooks to hook-classification.yaml, drop the 2 stale SparkDex entries (P1-1); raise coverage on the new surface (P1-2).


## R2 - Collapse to one manifest, derive denomination from amountRoles (P1-3). 

The contract is now authoritative; reconcile or delete the contradicting TS denom. Add a real git diff --exit-code drift gate and wire whichever validator survives into CI (P2-3).



## R3 - Bugfix 

Fix getOutToken for the 13 loan/firelight hooks (P2-1) and the MorphoWithdraw description (P2-2).



## R4 - Add invariants 

Add the all-hooks conformance invariant asserted from hookType() (P2-4, closes SUP-19991 properly).



## R5 - Remove unnecessary information 

Strip .claude/** incl. the Lotus report; add a PR description (P2-6).


## R6 - Sequence the deploy 

Coordinate v2-core + OMS ABI regen (array-aware M2/M3) + SuperBundler in lockstep; verify the bundler repo (P2-5).


## R7 - Scope (carried from 12-June item 6) 

If batching is audit-driven, OK, but stacked commits separating interface-standardization from net-new hooks (RFLR V2/V3, OpenOcean) would make the audit diff far cleaner.




# Bottom Line 

The design is right and most of my T4 asks landed — including the on-chain-denomination move 

Request changes on the red CI, the manifest duplication/contradiction, the outToken bug, and the hygiene; none threaten the architecture.

NicolaBernini (changes requested):
# Overview 
Overall it's good work, just a few blocking things, find the details here 
https://github.com/superform-xyz/v2-core/pull/923#issuecomment-4842427138

supervaulter:
| # | Finding | Status | Notes |                                                                                                                                                           
  |---|---|---|---|                                                                                                                                                                          
  | **P1-1** | Stale SparkDex entries in sizing manifest | ✅ Done | Removed `SwapOpenOceanSparkDexHook` + `ApproveAndSwapOpenOceanSparkDexHook` OVERRIDES; new OpenOcean keys auto-detected                                                                                                                                                                                    
  | **P1-2** | Coverage gate RED — no `getOutToken` tests on real loan hooks | ✅ Done | 14 new unit tests across Morpho (7), AaveV4 (6), Firelight (1); all 14 pass |                       
  | **P1-3** | Dual-manifest `denom` contradiction (TS vs Python manifests) | ✅ Done | Removed `denom` from `generate-hook-sizing-manifest.ts` and validator; `manifests/hooks.json` is sole
   authority |                                                                                                                                                                               
  | **P2-1** | `getOutToken` returning `address(0)` for 14 loan hooks | ✅ Done | Fixed `_setOutToken` calls added to `_postExecute` in all Morpho + AaveV4 loan hooks |                     
  | **P2-2** | `MorphoWithdrawHook` metadata mismatch | ✅ Done | Corrected hook classification YAML entry |                                                                                 
  | **P2-3** | No drift gates in CI | ✅ Done | Added drift-gate job to `contracts-ci.yml` |
  | **P2-4** | No universal sizing-interface conformance check | ✅ Done | Added P3 invariant (`lint_sizing_invariant`) to `lint_hook_manifest.py` |
  | **P2-5** | OMS ABI needs regen + SuperBundler array-aware encoder | ⚠️  Deferred | Requires changes to two off-chain services (OMS, SuperBundler) that live outside this repo. The
  contract ABI changes are done here; the consumer side must be coordinated separately by the team owning those services. No code in v2-core can unblock this. |
  | **P2-6** | `.claude/` session files committed to git | ✅ Done | `git rm --cached -r .claude/`; added `.claude/` to `.gitignore` |
  | **P3-1** | On-chain `name`/`description` not surfaced in manifest | ⚠️  Deferred | Every hook exposes `name()` and `description()` on-chain. Surfacing them in `manifests/hooks.json`
  would require either (a) a live RPC call during generation — fragile and slow — or (b) duplicating the strings into `hook-classification.yaml` — a maintenance burden. Neither approach has
   a clear benefit over reading the hook contract directly. Needs a product decision before implementation. |
  | **P3-2** | No explicit `sizeless` convention in manifest | ⚠️  Deferred | Sizeless hooks are currently tracked via an ad-hoc `sizelessHooks` list in `hook-enrichment.yaml`. Formalising
  this as a first-class manifest field requires agreeing on the right schema shape (boolean flag, separate section, or null `mode`) and updating every consumer. Low urgency; the current
  list works and is validated by CI. |
  | **P3-3** | External hook sizing vs sizeless ambiguity | ⚠️  Deferred | Some hooks (e.g. bridge hooks) delegate amount determination to an external contract so they have no splicing
  position, but they are not "sizeless" in the same sense as config or permission hooks. Distinguishing these two categories would add a new axis (`reason: external \| unconditional`) that
  no consumer currently needs. Revisit if OMS or the bundler needs to route differently based on this distinction. |
  | **P3-4** | `NativeTransferHook` had no key constant in `Constants.sol` | ✅ Done | Added `NATIVE_TRANSFER_HOOK_KEY = "NativeTransferHook"`; hook now auto-detected at `replaceCalldata /
  AMOUNT_POSITION=20` |
  | **P3-5** | `Denomination.UNITS` not modelled | ⚠️  Deferred | `MorphoLendHook` sets `outAmount` in supply shares, not underlying asset units. The existing denomination field only
  distinguishes `TOKEN` vs `ASSETS`; there is no `SHARES` value. Adding it requires auditing every hook's `outAmount` semantic, updating the enrichment YAML, and teaching OMS/SuperBundler
  to handle the new value. The risk of silent mis-accounting means this needs careful cross-team design rather than a quick patch. |
  | **P3-6** | Dead `tooling/enrich_manifest.py` (368 lines, no callers) | ✅ Done | Deleted; enrichment logic had already been folded into `generate_hook_manifest.py` |

NicolaBernini:
# Overview 

Overall good progress but still requires some fixes 

# Details 

## S1 


### S1.1 - lint-manifest RED 

The new drift gate (P2-3) trips on non-determinism, not real drift. 
The generator emits the per-chain addresses map in unstable key order: CI's regenerated file has "14","56","42161" where committed has "56","42161","14" — identical addresses, reordered. Gate is correct in spirit but flaky. Fix: deterministic key sort in generate_hook_manifest.py (chain IDs / address maps), then regenerate-and-commit. (Same class as the generatedAt exclusion he already added in 437e7db2.)



### S1.2 - coverage RED "Stack too deep" compiler error, not a coverage threshold. 

Under forge coverage --ir-minimum, a swap/router hook overflows 16 slots: Variable …_mpos is 1 too deep [ … var_tokenIn var_tokenOut var_deadline … var_account var_prevHook … ]. build/test pass (via-IR); the added name()/description()/amountRoles/getOutToken locals deepened the stack. The 14 new tests can't compile under coverage. Fix: block-scope locals / extract helper in the offending swap hook, or enable via_ir for the coverage profile.


## S2 

### S2.1 - Factual slip — P3-5 deferral note. 

The "there is no SHARES value" in the denomination field. The on-chain Denomination enum is {TOKEN, ASSETS, SHARES} (ISuperHook.sol:121-125) and SHARES is actively used (Redeem4626). He's conflating amountRoles input-denomination (has SHARES) with MorphoLendHook.outAmount output semantics (no denom field). Deferring UNITS is fine; the stated reason is wrong.


### S2.2 - P2-4 caveat. 

The lint_sizing_invariant is manifest-level (reads erc165 synthesized from YAML) — catches a missing YAML entry, NOT a .sol that forgot _supportsSizingInterface()=true. Pair with an on-chain Foundry invariant (iterate hooks → call supportsInterface) to fully close SUP-19991.

NicolaBernini:
# Overview 

Very close, just a couple of major findings, the other ones are minor 

I'd recommend fixing all the P1s and possibly P2s 


# Findings

## Important - P1

### P1.N1 · ClaimFailedTransferHook — getOutToken() returns address(0)

src/hooks/claim/stargate/ClaimFailedTransferHook.sol:165 calls _setOutToken(asset, account) but asset (BaseHook transient, defaults 0) is never assigned — only reference in the file is that line. Single-token claim with a real non-zero _setOutAmount(delta) → a downstream PASSTHROUGH hook forwards (token=0, amount=realDelta). Same class as the P2-1 loan bug Cosmin fixed; this hook was missed. Fix: _setOutToken(BytesLib.toAddress(data, <tokenOffset>), account).



### P1.N2 · ClaimFailedTransferHook — 52-byte header migration is COMMENT-ONLY (false green)

The 112ad5f7 diff added @notice placeholder = slice(0,52) to the doc, but the code was not migrated: still reads adapter@0, token@20, amount@40, MIN_DATA_LENGTH=72, AMOUNT_POSITION=40; decodeAmounts/amountRoles/replaceCalldataAmounts all operate on pre-header offsets. Every sibling claim hook reads ≥52 (Merkl feeReceiver@52; Fluid/Gearbox/Yearn asset=toAddress(data,52)). Systemic scan: this is the ONLY hook in the repo reading real fields at 0/20/40.

The unit test is ALSO un-migrated: _encodeData(adapter, token, amount) = 72 bytes, NO header, asserts adapter@0 — so it passes and gives false assurance. Contrast the migrated WrappedNativeHook test: abi.encodePacked(bytes32(0), address(0), amount, wrap, usePrev) = 52-byte header + fields at 52+.

Impact: if the strategy/OMS prepends the 52-byte header uniformly (as the standard mandates and WrappedNativeHook's own test encodes), this hook mis-decodes in production (adapter read from header zero-bytes → ADDRESS_NOT_VALID revert / wrong claim). If it's an intentional header exception, the doc is wrong. Reconcile — migrate hook + test, or revert the doc. N1 & N2 interact: if header is applied, the correct token offset becomes 72.


## P2 - Quite Important 

### P2.N3 · WrappedNativeHook — unwrap direction reports wrong outToken

_postExecute sets _setOutToken(WRAPPED_NATIVE, account) unconditionally. For unwrap (WETH→ETH) the output is native, which the codebase represents as address(0) (cf. Swap1InchHook.sol:413 if (dstToken == NATIVE || dstToken == address(0))). A PASSTHROUGH hook chained after an unwrap is told it holds WETH the account just burned. Tests assert getOutAmount after unwrap but never assert getOutToken (uncovered). Fix: _setOutToken(wrap ? WRAPPED_NATIVE : address(0), account).



### P2.N4 · MerklClaimRewardHook — getOutToken() == address(0) (misleading, low-harm)

MerklClaimRewardHook.sol:171 _setOutToken(asset, ...), asset never assigned. Defensible because Merkl is inherently multi-token (claims params.tokens[]) so there's no single out-token, AND _setOutAmount(0) → downstream PASSTHROUGH forwards (0,0) (harmless). But it references an unassigned asset exactly like the P2-1 bug. Fix: drop the _setOutToken call, or set explicitly with a comment "multi-token claim → no single outToken."


## P3 - Bonus 

### P3.N5 · WrappedNativeHook — MIN_DATA_LENGTH defined but unused

Constant at line 47 is never referenced — no explicit data.length validation (relies on BytesLib bounds). Cosmetic; ClaimFailedTransferHook does use its own MIN_DATA_LENGTH for a DATA_NOT_VALID revert.

NicolaBernini:
# Overview 

OK looks mergeable 

Just a few of non-blocking comments, better to address them in a separate PR 


# Findings 

## P1 

### P1.S1 (non-blocking, TRACK) — swap standardization is PARTIAL; prod-deployed stragglers keep divergent layouts

Not migrated to ISuperHookSwap/SwapCalldataLayout, yet deployed in prod/staging:

| Hook | Deployed | AMOUNT_POS | USE_PREV | Layer1? | 
|---|---|---|---|---|
| **SwapUniswapV3Router02Hook** / ApproveAnd… (IV3SwapRouter) | **Base prod+staging** | 128 | 192 | ❌ | 
| **SwapUniswapV2Hook** / ApproveAnd… | **Flare prod+staging** | 124 | 188 | ❌ | 
| SwapSparkPSMExactIn/Out / ApproveAnd… | Base prod+staging | 92 | 208 | ❌ (amount matches, useprev differs) | 
| SwapUniswapV4Hook | Base staging | 172 | — | ❌ | 
| pendle (RouterSwap/Redeem/Unified), spectra | — | bespoke | — | ❌ | 


UniV3 subtlety: the migrated SwapUniswapV3Hook wraps ISwapRouter (router01); prod-Base actually deploys the Router02 (IV3SwapRouter) contract, which this PR header-migrated but did NOT move to Layer1. So Base's live UniV3 hook stays on the bespoke layout (amount@128), and the standardized variant isn't the one deployed. Can't just swap addresses — different router interface.

Risk / why it's non-blocking: these are already-deployed and work under the current per-hook (manifest/on-chain) sizing regime; merging #923 does NOT redeploy or break them. The hazard materialises only when Superman/superbundler switches to encoding SWAP calldata uniformly via SwapCalldataLayout while still pointing at un-migrated addresses → mis-decoded amount/useprev/slippage. So: safe to merge; two follow-ups before the OMS uniform-encoder flip (M3): (a) confirm the swap-encoding path stays strictly per-hook, or (b) migrate the prod stragglers — priority UniV3 Router02 (Base) + UniV2 (Flare) — or deprecate them in favour of the migrated variants and repoint registry/deploy. Consistent with the M2/M3 multi-repo lockstep concern from the original review.


## P3 - Nits 

### P3.S2 (nit) — on-chain sizing invariant still hand-enumerated

test/unit/hooks/HookSizingInterface.t.sol explicitly asserts supportsInterface per named hook (transferERC20, swapUniV3, deposit4626, morphoSupply, aaveSupply, merklClaim, …) — comprehensive but not a registry-wide loop, so a newly-added hook omitted from the list won't be caught. Same as prior P2-4 caveat. Cosmetic.

NicolaBernini (approved):
# Overview 
PR is mergeable in this state, just a couple of things, for details see here 
https://github.com/superform-xyz/v2-core/pull/923#issuecomment-4868613184

The S1 specifically is worth getting a look but OK to address it in a separate PR since this one is already big enough

vikramarun:
Cosmin — I did another pass focused only on the contract-related SuperformOS 2.0 substrate in this PR. This is not meant to say the full OS 2.0 surface is covered here; it’s specifically the hook/manifest/ABI layer. A few things I’d double-check before treating these artifacts as safe for downstream consumers:

## Must double-check

1. **Supply-and-borrow output semantics may not match consumer expectations**

For the compound loan hooks, please double-check whether the standardized “output” is meant to be the borrowed asset or the collateral delta.

Examples I’d re-verify:
- `src/hooks/loan/aave-v4/AaveV4SupplyAndBorrowHook.sol`
- `src/hooks/loan/morpho/MorphoSupplyAndBorrowHook.sol`

From inspection, these flows borrow a loan asset, but `_postExecute()` appears to set `outAmount` / `outToken` around the collateral token/delta. At the same time, `amountRoles()` / manifest metadata can make consumers treat the borrowed amount as `OUT`.

If intentional, this should be explicitly documented as “collateral consumed / accounting output” and not assumed to be the pipeable downstream token. If not intentional, `getOutAmount()` / `getOutToken()` should probably reflect the borrowed loan token/amount, or these hooks should be marked non-pipeable / special-case in the manifest.

The important invariant for OS consumers is: a downstream hook using previous-hook output should not accidentally size against collateral when the account actually received the borrowed token.

2. **Morpho compound hook arity looks worth rechecking against the public manifest**

Please verify the arity/semantics line up between:
- on-chain `decodeAmounts()` / `amountRoles()`
- `manifests/hooks.json`
- `hook-sizing-manifest.json`

The Morpho compound hooks look especially easy to misread because the loan amount can be derived internally while the public manifest can still describe multiple semantic roles. If the manifest advertises two semantic slots but the runtime sizing ABI exposes one amount slot, downstream resolver/UI code can drift.

3. **Cross-manifest parity should probably be a CI failure, not a manual convention**

The new generated artifacts are useful, but they are not generated from one single source of truth:
- `manifests/hooks.json`
- `hook-sizing-manifest.json`
- constants / classification YAML / source contracts

Current CI appears to validate generated-vs-committed parity for each artifact, but not full parity across the artifacts.

I’d add a cross-manifest linter that fails on:
- hook exists in one manifest but not the other, unless explicitly allowlisted
- name/key drift such as old constants vs renamed hooks
- `amountMeta.length` not matching sizing slots / `decodeAmounts()` arity
- `sized=false` disagreeing with sizing mode semantics (`external`, `sizeless`, etc.)
- pipe mode / amount role mismatches that would confuse a resolver

This matters because SuperformOS consumers will treat these manifests as a shared registry. If the registries disagree, the failure moves from CI into Superman / bundler / backend behavior.

4. **Sizing warnings should be intentional and allowlisted**

I’d be cautious with “validation passed with warnings” for calldata sizing. Manual or inherited offsets are exactly where stale ABI layouts can silently corrupt calldata.

Please double-check any current warnings around:
- unverifiable manual offsets
- passthrough hooks that still have offset / `replaceCalldata` sizing modes
- inherited/inlined layouts that the validator cannot prove automatically

My strong preference: make these fatal unless explicitly allowlisted with a short rationale. For high-risk offsets, add round-trip tests proving `decodeAmounts()` and `replaceCalldataAmounts()` touch the same slots.

5. **The sizing interface split is a consumer footgun**

`ISuperHookOutflow.replaceCalldataAmounts(bytes,uint256[])` says amounts must match `decodeAmounts`, but `decodeAmounts` lives on `ISuperHookInflowOutflow`, not on `ISuperHookOutflow` itself.

Also, `BaseHook.supportsInterface()` appears to map both sizing/decode support and outflow replacement support through the same `_supportsSizingInterface()` path, with special-case overrides needed for decode-only hooks.

Please double-check that every hook advertises ERC165 support exactly for what it implements. A future decode-only hook should not accidentally advertise `ISuperHookOutflow` and invite consumers to call `replaceCalldataAmounts()`.

Potential fixes:
- make `ISuperHookOutflow` inherit the decode interface, or
- introduce separate internal support booleans for decode/roles vs replacement, or
- add runtime tests that all hooks advertising `ISuperHookOutflow` actually implement replacement and round-trip correctly.

6. **Runtime ABI validation would make the manifest much safer**

The manifest checks are good, but I’d add a small runtime/ABI validation layer that deploys or instantiates representative hooks and verifies:
- hooks marked sized support the expected ERC165 interface(s)
- `decodeAmounts()` / `amountRoles()` are callable where advertised
- `replaceCalldataAmounts()` is only advertised where implemented
- replacement round-trips preserve all non-amount calldata fields
- manifest amount roles match runtime roles

That gives Superman / bundler / backend a stronger guarantee than “the JSON regenerated cleanly.”

## Product framing / scope note

This PR is valuable as a contracts standardization layer: hook names/descriptions, amount metadata, sizing, pipe modes, and generated manifests. I would not frame it as “SuperformOS 2.0 complete” on its own. The remaining OS surface still needs the resolver/registry/readiness layer that turns this metadata into safe product behavior across Superman, backend, bundler, and app consumers.

Net: strong direction, but I’d lock down the output-token invariants, manifest parity, and ERC165/sizing guarantees before downstream systems treat these manifests as authoritative.

PR #934, opened 2026-06-26
feat: morpho blue oracle and tests

supervaulter:


NicolaBernini:
# Overview

Overall good but there are a few blocking issues 

I'd say P1 are required while the test is option but P2 would indeed be good to have 



# Findings 

## P1 Findings

### P1.1 — PPS quantization / interface breach for low-decimal loan tokens.

IYieldSourceOracle.decimals() NatSpec: "the number of decimals used by the yield source's share token". Morpho supply shares carry the SharesMathLib virtual offset (VIRTUAL_SHARES = 1e6), i.e. effectively loanDecimals+6. The PR returns loanToken.decimals() and prices 10**loanDec shares in getPricePerShare. For a USDC market: pps_raw = toAssetsDown(1e6 shares) ≈ 1 — the author knows (test comment: "PPS for a 6-decimal token is ~1 atomic unit") but the consequence is unhandled. BaseLedger computes costBasis += shares·pps/10^dec on inflow and profit = assets − costBasis on outflow: with pps floored at 1 while real pps is e.g. 1.05, cost basis is understated ~4.8% → phantom profit → performance fee overcharge on principal; error grows toward 50% as real pps → 2. Mitigated today only by the interview-notes decision "standalone, do NOT register in SuperLedgerConfiguration" — nothing enforces that, and even monitoring PPS is useless when it moves in +100% quanta.
Fix: decimals() → loanDec + 6; getPricePerShare() → toAssetsDown(10**(loanDec+6)) (≈10^loanDec at price 1.0, full precision; empty-market case returns exactly 10^loanDec). Ledger math stays consistent; matches MetaMorpho's DECIMALS_OFFSET philosophy. Same flaw exists in the merged periphery oracle → follow-up ticket there regardless of P1.2's outcome.



### P1.2 — Unacknowledged duplication with v2-periphery MorphoLendYieldSourceOracle (merged PR #302, 8 Apr 2026).

Same scope (Morpho Blue supply-side yield source oracle, in-view accrual). The PR's own research/repo-analysis.md never mentions it — the research pass analyzed v2-core only. Differences worth a conscious decision:

Market identity: periphery = manager-chosen arbitrary yieldSourceId address; core = deterministic pseudo-address address(uint160(uint256(marketId))) (lower 20 bytes of market id). Two different keys for the same market across the stack (deployment scripts, superman ys config, monitoring).

Accrual math: periphery fee-share mint omits the − feeAmount denominator (deviates from Morpho, slightly overstates PPS on fee-enabled markets); core matches Morpho/MorphoBalancesLib exactly.

Robustness: periphery = no elapsed cap, silent try/catch fallback to stale state on IRM failure; core = 365d cap, IRM whitelist, revert on unregistered.
Recommendation: consolidate on the core implementation (it is strictly better), deprecate the periphery one, and check whether periphery's was ever deployed/registered anywhere.



### P2 Findings

### P2.1 — All committed spec docs describe the abandoned design. 

The spec.md, technical-spec.md, interview-notes.md (~1000 lines) specify a permissionless MorphoBlueMarketWrapper per market; the shipped code is a permissioned registry — the trust model flipped between spec and implementation and none of the docs were updated (the session file itself records "MorphoBlueMarketWrapper.sol — replaced by registry"). Committed specs that contradict the code are worse than no specs.



### P2.2 — The security report's P0 is a false positive and the report was never updated post-fix. 

The P0 claims a read-only-reentrancy window where "totalSupplyAssets has been incremented but totalSupplyShares has not" — Morpho Blue updates both totals and the position before any token transfer (CEI), so that window does not exist for supply/withdraw. The implemented mitigation (IRM whitelist) actually addresses P1-01 (rogue IRM), which is real. Also P2-06 (flash-loan PPS inflation) is disproven by the PR's own fuzz test (testFuzz_supply_doesNotChangePPS). The report still says "Verdict: FAIL — 5 blocking findings" even though 4-5 of them are fixed in this very commit — resolution status should be recorded or the report will mislead auditors.



### P2.3 — Spec-mandated bit-exact parity test is missing. 

Interview notes Round 3: "Bit-exact parity required… validated by fork tests at specific block numbers." No test calls Morpho.accrueInterest(mp) on-chain and compares the resulting stored market() state against the oracle's prior view output. Existing tests assert only monotonicity/approximation. This is THE test that justifies view-replication; it's ~15 lines.



## P3 Findings

### P3.1 — "Hook lifecycle" tests hand-replicate the build() batch 

(comment: "Replicate the executions built by lendHook.build()") instead of executing build()'s actual output or going through SuperExecutor (_getExecOps/executeOp like other integration tests). If build() drifts, tests keep passing.



### P3.2 — Post-deregistration the oracle reverts for that key; 

if ever ledger-registered, fee-charging outflow accounting for held positions bricks until re-registration. Timelock helps; add an ops-runbook check ("verify no active Superform positions before deregistering").



### P3.3 — .claude/sessions/context_session_18.md committed (internal AI session context, itself describing the outdated wrapper design) 

Same hygiene issue as PR #923's .claude/Lotus files. Gitignore it.



### P3.4 PR hygiene 

Empty body, no Linear ticket, spec marked "Pod leader approval pending", and two unrelated test-flakiness fixes bundled (Pendle Odos try/catch+vm.skip, Spark PSM deal() liquidity top-up — both individually fine and well-commented, but belong in a chore PR).

NicolaBernini (changes requested):
# Overview 
Overall it's good, just a few blocking issues, see details here
https://github.com/superform-xyz/v2-core/pull/934#issuecomment-4866741679

vikramarun:
Cosmin — I did a pass specifically from the Morpho/equities-launch angle. The core oracle direction looks good, and the low-decimal PPS issue looks addressed now with `loanDecimals + 6`.

A few things I’d double-check / make explicit before we treat this as launch-ready for tokenized equities / Morpho markets:

1. **Confirm the accounting unit for equities markets**

This oracle values Morpho supply shares in the market `loanToken`.

That is correct if Superform accounting wants loan-token-denominated supply-side value. But for the equities launch, please explicitly confirm this is the intended unit and that no downstream surface expects equity collateral value, USD NAV, or the Morpho market oracle price.

In other words: this is a Morpho supply-position oracle, not an equity/NAV oracle. That distinction should be clear in docs/config before launch.

2. **Track the launch wiring separately from this contract PR**

This PR gives us the oracle/registry substrate, but the actual equities launch still depends on configuration work outside this diff.

Can you make sure the launch checklist is tracked somewhere? Specifically:

- registry deployment per target chain
- oracle deployment
- approved IRMs
- exact market registrations once the Morpho markets are known
- `SuperLedgerConfiguration` oracle id / fee config
- ownership / role transfer plan
- monitoring for registered market keys

Not saying all of that needs to be in this PR — just want to avoid merging the substrate and accidentally treating the launch/config layer as done.

3. **Deregistration can still break live reads**

The 2-day timelock is good, but after `executeDeregisterMarket()` the oracle reverts for that market key. If there are active Superform positions, that can brick PPS/TVL/fee-preview reads until re-registration or migration.

I’d document/enforce an ops invariant: do not deregister a market with active Superform accounting positions unless it has been migrated/deprecated first.

4. **Add one bit-exact Morpho accrual parity test**

The fork tests are broad, but I’d still like one direct parity test:

- compute the oracle’s view-accrued state
- call Morpho `accrueInterest` on the same market
- compare the resulting stored Morpho market state against the oracle’s prior computed values

That’s the strongest proof that the view-replayed accrual matches Morpho exactly, especially before using this for launch accounting.

5. **Clean stale specs before merge**

A lot of the committed spec/research docs still describe the old `MorphoBlueMarketWrapper` / permissionless-wrapper design, while the implementation is now a permissioned registry.

That’s a different trust model. Please update or remove stale docs so future auditors/agents don’t reason from the wrong architecture.

Net: the implementation looks directionally right, but I’d separate “oracle code exists” from “equities launch is wired and safe.” The latter needs explicit config/runbook tracking.

NicolaBernini (approved):
# Overview
Approving to be able to move on 
As discussed with @supervaulter additional feedback can be tackled in #942

PR #936, opened 2026-06-29
fix: safe staticcall for ClaimRFLRV2Hook new-account balance snapshot

supervaulter:
Fixes OPERATION_FAILED on first-time rFLR claims where no RNatAccount exists yet. Replace direct IERC20.balanceOf with _snapshotRNatBalance returning (bool ok, uint256 balance) to distinguish no-account (!ok) from existing-account-zero-balance (ok && 0). _preExecute skips _setOutAmount when !ok; _postExecute computes the full claimed delta.


ronnypanford (approved):
Correctly fixes the issue. and the delta calc is good.

PR #942, opened 2026-07-03
feat: swappers layer1 and 2 params + abi.decode and encode for payload + tests

supervaulter:

PR #944, opened 2026-07-09
feat: uni v3 lp oracle

supervaulter:


NicolaBernini:
# Overview 

The architecture seems right: a singleton oracle backed by a UniV3CLPRegistry (positionKey pseudo-address → precomputed config, mirroring the MorphoBlueMarketRegistry pattern from #934), pricing liquidity from a Chainlink-derived sqrtPriceX96 instead of pool slot0 (flash-loan resistant, the gold standard), with the decimals()=18 / 1 share = 1e18 liquidity units convention correctly coherent with BaseLedger (the #934 decimals lesson applied). 

But request changes on two P1s, both proven on-chain today: 
P1.1 Aerodrome Slipstream support is broken — the fee() try/catch heuristic assumes Slipstream pools lack fee(); they don't, so every Slipstream position fails the match and getTVLByOwnerOfShares returns 0 on Base, the launch chain; 
P1.2 the 60s inter-feed skew check bricks getPricePerShare for real feed pairs — measured live skew between Base ETH/USD and USDC/USD is 19.2 hours. 

Both P1s were introduced in the final "octane fixes" commit; the fork tests mock around exactly these two failure modes.


# Findings 

## P1 (required fix)

### P1.1: Slipstream position matching is broken: fee() does NOT revert on Slipstream pools

The `_getPoolFeeOrTickSpacing` (oracle L361-367) resolves the expected NFT match value as: try pool.fee() → return it; catch → return tickSpacing(). 
The NatSpec (L17-18, L32) claims "Aerodrome Slipstream pools lack fee()" while they don't: Verified on-chain 2026-07-20 against the exact pool used in the Base fork tests (Slipstream WETH/USDC 0xb2cc224c1c9feE385f8ad6a55b4d94E92359DC59):

fee() (0xddca3f43) → 199 (dynamic fee in pips, set by Aerodrome's fee module, not even a static tier)

tickSpacing() (0xd0c93a7c) → 100

Slipstream NFT manager 0x8279…5b72 positions(0xa9dc) slot 5 → 100 = tickSpacing (Slipstream's NPM stores int24 tickSpacing where UniV3 stores uint24 fee)

So on Slipstream the try branch succeeds with 199, _matchPosition (L431-435) compares it against the NFT's 100 → no position ever matches → getBalanceOfOwner/getTVLByOwnerOfShares return 0 for every Aerodrome position, on the chain the whole ticket (SUP-20132, Base Equities Launch) targets. 
A SuperVault holding a Slipstream LP would carry it at zero.

**Why tests missed it**: 

the Base fork section (L860-998) only exercises the PPS path (registry+feeds); getBalanceOfOwner/TVL-by-owner are never tested against a real Slipstream NFT — the real-LP-holder scan test exists only for Ethereum UniV3.

**Recommended Fix**: don't infer live. 
Store the expected match value (or a dexType enum) in PositionConfig at registration — consistent with the registry's precompute-everything design. Note fee is mutable on Slipstream (fee module) and tickSpacing is immutable, so the correct per-DEX key is: UniV3 → pool.fee(), Slipstream → tickSpacing(); a live-read snapshot of fee() would be wrong on Slipstream even if the heuristic were inverted. Then add the missing fork test: real Slipstream NFT (or minted position) → getTVLByOwnerOfShares > 0.


### P1.2: MAX_FEED_SKEW = 60s bricks getPricePerShare for real Chainlink feed pairs

The `_getPrices` (L314-317) reverts STALE_PRICE when |updatedAt0 − updatedAt1| > 60. 
Chainlink feeds are deviation + heartbeat push feeds with asymmetric cadences: Base ETH/USD ≈ 0.15% dev / 20-min heartbeat; Base USDC/USD ≈ 0.3% dev / 24h heartbeat (USDC essentially never breaches deviation, so it updates ~once a day). 
Measured live at review time: ETH/USD updatedAt = 2026-07-20 09:37:33 UTC, USDC/USD = 2026-07-19 14:26:19 UTC → skew = 19.2 hours. 
The oracle would revert right now for the fork-test pool; feasibility of both feeds updating within 60s of each other is a few minutes per day at best. For the Base Equities Launch it's worse: tokenized-equity feeds update during market hours only (frozen nights/weekends), so equity/USDC positions would be unpriceable virtually always.

Blast radius: getPricePerShare is on the critical path of BaseLedger._updateAccounting (L282, reverts on failure) → any ledger-registered CLP flow reverts; and via the new batch try/catch, supervault-pricing (app/services/yield_source.py L134 uses getTVLByOwnerOfSharesMultiple) gets a silent 0 → the LP position vanishes from SuperVault NAV (see P2-1). 

The fork tests are the tell: helper _syncFeeds (L128-148) mocks both feeds to identical timestamps with the comment "At pinned fork blocks, real feeds may violate these" — i.e. the suite validates the check's mechanics while masking that it fails against reality. The 775-line edge suite likewise only proves the check behaves as coded, with author-controlled timestamps.

Why the check is conceptually misguided: for deviation-based feeds, an old updatedAt means the price hasn't moved beyond the deviation threshold — the cross-rate error from asynchronous updates is already bounded by deviation settings, and per-feed maxStaleness (per-position configurable, correctly) bounds the tail. 
The merged UniV2LPYieldSourceOracle uses the same two-feed cross-rate with no skew check — this PR is inconsistent with its own precedent. Fix: delete the check (my preference), or make it per-position config with values ≥ max(heartbeat0, heartbeat1) — at which point it's redundant with maxStaleness. This appears to be an over-tight response to Octane's "inter-feed skew" finding.


## P2 findings

### P2.1: Batch silent-zero semantics change is fail-open on the NAV path (all 18 oracles). 
The AbstractYieldSourceOracle try/catch (L176-185) converts any oracle failure (stale feed, sequencer down, deregistered position, out-of-gas inner call) into 0 with no signal — for every oracle inheriting the base, not just CLP. supervault-pricing consumes exactly this batch to value SuperVault yield sources: a transient Chainlink hiccup would silently shrink NAV by the position's weight → mispriced deposits/redeems. This is the same fail-open class as the 16 Jul EE zero-floor P0. Prefer: keep loud failure and let the off-chain caller degrade per-entry, or return a parallel success[] mask; at minimum the pricing service must treat 0-from-batch as "verify with a direct call before use". Also note the DoS protection is partial anyway (inner OOG still consumes 63/64 of gas).


### P2.2: Uncollected swap fees are excluded from valuation. 
The `_matchPosition` discards tokensOwed0/1 and the feeGrowth fields; PPS values principal liquidity only. Swap fees are the yield of LPing — the position's balance-sheet value is systematically understated between collects, then NAV jumps when a (future) collect hook realizes them; Octane's performance-fee-undercharge finding is the ledger-side face of the same gap. May be an acceptable v1 scope decision (conservative direction, and unpoked fee computation needs feeGrowthInside reads), but it must be an explicit, documented decision, and the fee-accounting story (how collected fees hit the ledger, if at all) must be settled before any SuperLedgerConfiguration registration.


### P2.3: Gauge-staked positions are invisible. 
Balance is enumeration of NFTs owned directly. On Aerodrome, LP NFTs are staked into CLGauges to earn AERO — the dominant yield; a staked NFT is owned by the gauge, so the oracle reports 0. The internal report downgraded this to "design limitation", but for the Base launch it's a strategy-design constraint that needs an explicit decision: either strategies never stake (forgoing most Aerodrome yield) or the oracle learns to read gauge stakes. Should be in NatSpec + strategy docs either way.


### P2.4: Enumeration is dust-griefable. 
Anyone can transfer worthless position NFTs to a SuperVault address; _getBalanceOfOwner is O(N) with 2 external calls per NFT, so a few thousand dust NFTs push the view past RPC eth_call gas caps → the TVL read fails → with P2-1, silently 0 → an attacker can knock a position out of NAV for the cost of minting NFTs on Base. Robust fix: value explicit tokenIds tracked by the strategy (it knows its own positions) instead of owner enumeration; or cap/paginate the scan.


### P2.5: No ground-truth accuracy test. 
Every fork assertion is > 0 / < 1e30 / monotonicity / round-trip ≤. There is no test comparing the oracle's valuation of a real position against the spot-derived truth (LiquidityAmounts at pool slot0, tolerance for CL-vs-spot deviation) — the one test that justifies the pricing approach (same class as the missing bit-exact parity test in #934). It would also have caught P1-A had it been written for Base.



### P3 / minor

### P3.1 Unchecked uint160 downcast of the Chainlink-derived sqrtPrice (L214) before clamping 
Extreme cross-rates (exotic decimals/price ratios) could wrap and clamp to the wrong boundary. Clamp in uint256 first, then cast.


### P3.2 refreshCircuitBreakerBounds doesn't refresh feed decimal scales 
An aggregator swap behind the proxy can change decimals(); bounds are refreshable but feedXScale isn't (Octane warning only half-addressed). One line to add.


### P3.3 Asset denomination is token0 by address sort. 
PPS/TVL are token0-denominated; nothing validates a ledger config's asset == token0, and for equity/USDC pools token0 could be either side depending on addresses. Ops/config footgun — document + validate at registration or config time.


### P3.4 getTVL()=0 
This is honest and documented, but consumers (getTVLMultiple, dashboards) can't distinguish "disabled" from "empty". 
Acceptable; worth a mention in integration docs.


### P3.5 Security report is stale vs final code 
Its findings (P2-1…P2-5, P3-1) were all fixed in cf9fc836+, but the committed report still lists them as open with pre-fix line numbers; the two post-report P1s introduced by the octane-fixes commit are obviously absent. Regenerate against head (the #934 lesson: don't let auditors internalize a stale report).


### P3.6 Hygiene 
PR body empty, no Linear ref (SUP-20132); no design spec committed (only the security report); KyberSwap vm.skip + Odos abi.encode payload change are unrelated scope (the latter confirms the #942 encoding-divergence item — worth cross-referencing there).

NicolaBernini (changes requested):
# Overview
Good start: architecture looks correct, there are just a couple of P1 issues requiring fix and would be better to address P2 as well, P3 are optional, see details here
https://github.com/superform-xyz/v2-core/pull/944#issuecomment-5020988535

NicolaBernini:
# Overview - Round 2 

OK P1.1 and P1.2 are fixed so LGTM 

Just a few comments 

New Round-2 findings

## P2 

### P2.1 (P2, pre-merge): the new Slipstream regression test is vacuous, proven on-chain 20 Jul 2026. 
The `test_base_realSlipstreamHolder_getBalanceOfOwner` scans NFT tokenIds 1–500 for a full-range (−887200/887200) WETH/USDC tickSpacing-100 position and silently passes if none is found ("test is inconclusive" comment). 
Probed the Slipstream NPM (0x8279…5b72) via batched eth_call: tokenIds 1–500 contain exactly 25 WETH/USDC ts=100 positions, all concentrated around tick −196xxx, none full-range (and all with liquidity 0 today). Since tick ranges are immutable after mint, the scan finds nothing at any pinned block → foundHolder == address(0) → the getBalanceOfOwner > 0 assertions never execute. This is the exact "coverage that isn't" pathology that let P1.1 through in round 1, reproduced inside its own regression test. Fix (small): mint a position in the test (deal WETH/USDC → approve → NPM.mint full-range) — deterministic, no scan; or minimally assertTrue(foundHolder != address(0)) / vm.skip so inconclusive shows as skipped, not passed. Note the fix itself is still adequately verified by unit tests with a mocked NPM storing tickSpacing — this is a test-quality gap, not a code defect.


### P2.2 (P2, pre-merge): generated-bytecode manifests are stale vs head. 
The script/generated-bytecode/UniV3CLPRegistry.json at 5cd4634e still contains refreshCircuitBreakerBounds (renamed) and the old 11-arg registerPosition without feeOrTickSpacing (last touched by the dev-merge commit, before the feedback commit). Registering a Slipstream position from this artifact would use the wrong ABI. CI is green despite this, so the manifest lint step (which Cosmin "fixed" per Slack) does not catch src↔artifact drift. Same regen-before-merge condition as PRs #941/#942, but here the changed function is the operational entry point.


## P3 

### P3.1: feeOrTickSpacing is stored unvalidated at registration (registry L225). 
An operator typo (e.g. passing fee 199 instead of tickSpacing 100 for a Slipstream pool, or 500 vs 3000 for UniV3) silently zeroes every balance/TVL read — the same failure surface as P1.1, moved from code to ops. Cheap in-pattern sanity check: require the value equals pool.fee() (try) or uint24(int24(pool.tickSpacing())) (catch/or). The registry already validates token0/token1 against the pool; this completes the set.


### P3.2: interface break rollout hazard. 
The getTVLByOwnerOfSharesMultiple signature change is consumed off-chain only (no on-chain callers in v2-core or any local Superform repo). Decoding the new two-output return with the old ABI happens to work (first head word offset still points at userTvls) — which is why nothing breaks silently today — but decoding old deployed oracles with the new ABI mis-decodes. Until all 18 oracles are redeployed, the fleet is mixed-signature; supervault-pricing needs per-oracle ABI awareness or the redeploy must be batched with the service upgrade.

NicolaBernini (approved):
# Overview 
OK the P1.1 and P1.2 from the previous review round have been correctly solved, good job, so now it's mergable, there are just a few P2 and P3 findings, see details here 
https://github.com/superform-xyz/v2-core/pull/944#issuecomment-5022946283

PR #961, opened 2026-07-23
feat: added paymaster new version

supervaulter:


ronnypanford (commented):
hey - looked at this vs how OMS actually submits UserOps today

**blocker:** the calldata checks don't match our path

OMS is:
- `userOp.sender` = SuperVaultExecutor (`0x183e...`)
- `callData` = `executeFromEntryPoint(strategy, args)` (`0x4fb2fbd5`)

this PR checks for Nexus.execute -> SuperExecutor.execute instead

so all our real strategy intents would fail `INVALID_CALLDATA` as written

**also missing:** strategy isolation - we should require the strategy in calldata == strategy in `paymasterAndData`. otherwise a session on vault A can still burn vault B's gas budget

what we need for OMS:
1. `userOp.sender == SuperVaultExecutor`
2. selector = `executeFromEntryPoint`
3. strategy arg matches `paymasterAndData` strategy
4. keep the 4M gas cap - that part is good

side notes:
- `DEFAULT_ALLOWED_SENDER` is hardcoded to prod SVE but used as calldata target, not sender. comment says `setAllowedSender` overrides the target - it doesn't
- staging SVE is `0x7AC580...`, so that constant breaks staging if we keep it like this

gas cap / default-deny `allowedSender` direction is right - just need the validation to match the actual OMS UserOp shape before we cut over

ronnypanford (commented):
looked at the new commit - this matches the OMS path now

checks look right:
1. `userOp.sender` must be the allowed sender (SVE once you set it)
2. selector = `executeFromEntryPoint` (`0x4fb2fbd5`)
3. calldata strategy == `paymasterAndData` strategy
4. 4M gas cap still there

nice - strategy isolation test is good too

one ops note: comment says `DEFAULT_ALLOWED_SENDER` means no setup call is needed, but validation still default-denies when `allowedSender[strategy] == 0`. so after `fundStrategy` you still need `setAllowedSender(strategy, SVE)` or nothing will sponsor. either auto-set on fund, or just fix the comment so we don't forget in deploy

nits:
- selector natspec says `(address, bytes)` but real sig is `(address, ExecuteArgs)` - selector is still correct
- staging SVE is `0x7AC580...` - fine since allowedSender is per-strategy, just don't copy the prod constant blindly

happy to approve for deploy after the allowedSender setup path is clear

ronnypanford (dismissed):
code side looks good now - the allowedSender comment is fixed and OMS validation path is still correct

blocker on the deploy manifests though:

`chore: deployed` set prod SuperSponsorshipPaymaster to `0x185f6f03abBe881A9f2018836202c8F6F0BCAFE2`

then `chore: fixed conflicts` reverted a bunch of prod `*-latest.json` (and `prod/latest.json`) back to the old `0xe043d74dc4eCDe253FdFE3725162ED0AcbC13c5C`

current split on this branch:
- still new `0x185f...`: Gnosis, Polygon, Sonic, Worldchain, Linea, Berachain
- reverted to old `0xe043...`: Ethereum, Base, Arbitrum, Optimism, BNB, Avalanche, Unichain, HyperEVM, Flare, Stable

staging looks consistent at `0x40491756bC208e0dfC10e84264B1f00473ff1721`

please restore `0x185f...` on the reverted prod files / `prod/latest.json` before merge, otherwise cutover will point main chains at the old paymaster

ops reminder still stands: after fund, `setAllowedSender(strategy, SVE)` (prod `0x183e...`, staging `0x7AC580...`)

nit only: selector natspec still says `(address, bytes)` instead of ExecuteArgs - not blocking

PR #971, opened 2026-07-29
chore: edits on PendlePTHook

supervaulter:


NicolaBernini:
# Overview 

Overall good start, a few findinga 

# Findings 

## P1 

### P1.1 LimitOrderData removal contradicts the 28 Jul agreement

The same 28 Jul thread that agreed the market-only inspect ALSO agreed OB support stays. Kurisu's scope split (p1785234055085839): in scope 
- T1 = AMM taker, 
- T2 = OB taker (market orders / filling resting orders); 

out of scope 
- T3 = placing limit orders, 
- T4 = LPing. 

Ronny confirmed: "limit orders are optional i.e OB, the functionality still exists" and "it is not affected".

T2 rides on LimitOrderData: the Pendle API returns an optimal AMM/OB split, and the OB fills arrive as normalFills/flashFills inside LimitOrderData. Hardcoding it empty = AMM-only execution → worse pricing and higher price impact on size (exactly what matters for SuperUSDC-scale PT trades).

Fix: Buy payload = abi.encode(ApproxParams, LimitOrderData); Sell payload = abi.encode(LimitOrderData); Redeem stays empty (par redemption, no venue choice). Restore _validateLimitOrders/_validateFillOrders/_validateOrder + their bounds/errors. Do NOT restore _validateSwapData (aggregator path stays out — that removal is agreed and consistent with the two-leg composition: external leg = Kyber/Odos hooks). Expiry-boundary note stays harmless: a sell-format payload routed to redeem at the expiry block is simply ignored (redeem never reads routingParams).


## P2 

### P2.1 Deployed to 16 networks pre-review; all deployments now invalid

Commit 6dd66bce "chore: deployed" (30 Jul 07:22 UTC) landed one minute before the Slack announce (09:23 CEST) and 25 minutes before Ronny's change request. PendlePTHook 0x265fc97E53FdB8816E9b6a67D59e592876cbDF8A (CREATE2, same address everywhere) is live on 11 prod chains (Ethereum, Base, Arbitrum, Optimism, BNB, Avalanche, Sonic, Berachain, HyperEVM, Unichain, +) and 5 staging chains.

Verified on-chain (30 Jul): runtime code on Base and Mainnet (13,268 bytes) exactly matches the PR's locked-bytecode deployedBytecode after masking the 19 immutable-reference slots, and the locked artifact's metadata source hash equals keccak256 of the PR-head PendlePTHook.sol (0x4670cbd7…). So the deploy pipeline itself was clean — bytecode, source, and chain state are mutually consistent. But restoring LimitOrderData changes the bytecode → full redeploy at a new address; the 0x265f… instances become orphans (make sure they are never SuperGovernor-whitelisted or registry-referenced).


## P3

### P3.1 manifests/hooks.json is stale relative to the deployment

The PendlePTHook manifest entry has addresses: {staging:{}, prod:{}} and availableChains: [] — it was generated 29 Jul 14:41 UTC (edits commit), before the deploy commit, and never regenerated. Same manifest-vs-reality drift class as PRs 941/942/944 (see [[project_pr941_deployment_review]] pattern). Moot for the addresses once the redeploy happens, but the regen step must land in the final PR state.


### P3.2 Why empty SwapData/pendleSwap are still constructed

They cannot be removed: pendleSwap and swapData are mandatory members of Pendle's TokenInput/TokenOutput ABI structs on IPendleRouterV4 — the calldata layout requires them. Zeroed values ARE the router's "skip" signal (SwapType.NONE → no external aggregator; empty LimitOrderData → AMM-only). The only latitude is cosmetic: explicit zero-literal (current, self-documenting) vs default-initialized SwapData memory (like the current LimitOrderData memory limit; line). Suggest answering Ronny with the ABI fact + optionally a tiny _emptySwapData() helper to deduplicate the three construction sites.


### P3.3 (gas): redeem path approves netTokenIn on both PT and YT (7 executions). 

The hook only routes to redeem when yt.isExpired(), and post-expiry redeemPyToToken burns PT only — the vault never holds YT in this design anyway. The 4 YT approve ops (grant+reset) look like dead weight; worth confirming against PendleRouter's post-expiry pull behavior and dropping if so.


## P4 

### P4.1 PR has no body/description; 968 also merged with none. 

For a contract shipping to 16 networks, the PR description should carry payload format + agreement links.

Positive: single readTokens() call passed down to builders; approve-reset-approve pattern consistent with repo idiom; outputMin==0 and scaled-min re-checks retained; .bak file cleanup.


# Checklist

Regenerate manifests/hooks.json with real addresses/chains.

SuperGovernor whitelist for the NEW address (Cosmin's 29 Jul GDM whitelisting round covered PendleUnifiedHook, not PendlePTHook).

erebor registry entry — none exists today (verified prod chain-1 registry 30 Jul: only PendleUnifiedHook + Record*AmortizedOracle hooks). inspect_params = yieldSource only; plus bundling_instructions chaining RecordPurchase/RecordRedemption as implicit sequence steps (Ronny confirmed 29 Jul the record hooks are invoked implicitly, analogous to 7540 claim legs; see fix plan in [[PT Buy Block I5 REBALANCE Failure — Debug (24 Jul 2026)]]).

OMS builder for the new payload shape (ApproxParams [+ LimitOrderData]) sourcing fills from the Pendle API split.

Merkle config re-propose per vault: one leaf per PT market (replaces per-output-token leaves).

NicolaBernini (changes requested):
# Overview 
Good start but need to fix P1.1, see details 
https://github.com/superform-xyz/v2-core/pull/971#issuecomment-5128635663

NicolaBernini:
# Second review 

Previous stuff has been fixed but there are some new findings 

# Findings 

## P1 

### P1.1 Sonic + HyperEVM prod NOT redeployed; manifest codifies the split-brain

The redeploy covered only 6 of the 8 prod chains. Verified on-chain 30 Jul: on Sonic (146) and HyperEVM (999), 0xb37C… has no code; the old AMM-only 0x265fc97E… (13,268 bytes, round-1 build) is still the only PendlePTHook there. The regenerated manifest doesn't flag this — it lists 0x265f… as the prod address for 146 and 999, so any consumer (erebor sync, OMS, registry seeding) will treat the old hook as canonical on those chains.

Why this fails silently, in the dangerous direction: the old buy path decodes abi.decode(payload, (ApproxParams)) — ApproxParams is a static 5-word struct, so a new-format payload (ApproxParams + LimitOrderData) decodes fine and the trailing limit-order bytes are ignored; the old sell path reads no payload at all. So new-format intents on 146/999 execute AMM-only with the OB fills silently dropped (only the outputMin floor protects the price). The reverse direction is loud (old-format payloads revert on the new contract), which means the 6 upgraded chains fail fast but the 2 stale chains degrade quietly.

Fix: run the deploy for 146 + 999, regen manifest, done. (The old hook deployed fine there in round 1, so pendleRouters config exists for both — the redeploy run just skipped them.)


## P2 

### P2.1 Aggregate latest.json stale for Berachain + stray .bak committed

Three artifacts now disagree about Berachain prod: per-chain output/prod/80094/Berachain-latest.json = 0xb37C (new), manifest 80094 = 0xb37C, but the aggregate output/prod/latest.json Berachain section = 0x265f (old). On-chain truth: 0xb37C is deployed on Berachain (verified) — so the aggregate is the stale one; its merge step half-failed, which is also how the stray prod/80094/Berachain-latest.json.bak (94 lines) ended up committed — ironic given commit 6dd66bce deleted the staging Base .bak. Fix the aggregate Berachain entry and drop the .bak. Meta-note: lint-manifest CI is green despite the per-chain/aggregate divergence — the linter checks schema, not cross-artifact consistency; worth a follow-up ticket.



## P3 Carried over from previous release, unchanged, not blockers but worth flagging them 

### P3.1 

Old orphan 0x265f… remains live on all originally-deployed chains (immutable) — after the 146/999 redeploy it must never be SuperGovernor-whitelisted or registry-referenced anywhere.


### P3.2 

Redeem path still approves netTokenIn on both PT and YT (post-expiry redeem burns PT only) — P3 gas nit.


### P3.3 

Downstream chain still ahead: SuperGovernor whitelist for 0xb37C (+ staging 0xCDBaC0), erebor registry entry (none exists — verified round 1), Record-hook bundling, OMS builder for the new payload, merkle re-propose per market. Ronny's ETA "PT support by end of week".

NicolaBernini (approved):
LGTM

PR #977, opened 2026-08-06
feat: deployed v2-core to RH & relay bridge

supervaulter:


NicolaBernini (commented):
Approving without reviewing, just deployoment

PR #982, opened 2026-08-11
feat: record pt hooks for PendlePTHook

supervaulter:


NicolaBernini:
# Overview 

The PR design looks correct but there is quite some operation burden attached to this, see findings 

# Details 

## Why the core mechanism is right 

PendlePTAmortizedOracleV2.recordPurchase(market, ptBought, twapDuration) (L165-208) does 
```
currentPtBalance = pt.balanceOf(msg.sender) 
```
after the buy and derives 
```
previousPtBalance = currentPtBalance − ptBought 

```
i.e. the recorded amount must equal the actual balance delta or the book-value math is wrong (or underflows). 
Manual amounts (the V1/V2 flow) could never guarantee that; TradeResult.outputAmount is literally that delta. 

Same for `recordRedemption` 

in

```
previousPtBalance = current + ptSold
```

And sySpent is computed on-chain: ptBought · getPtToSyRate(twapDuration) / 1e18 with twapDuration ≥ getMinTwapDuration(market) (default min 300) — Ronny's ask (1) satisfied with no off-chain input. 

Both call signatures verified: hook string → vendor IPendlePTAmortizedOracleV2 → real oracle (the unit-test mock implements the same vendor interface, so no mock-masking).

The redemption side closes a real accounting bug: old hooks' usePrevHookAmount read getOutAmount = asset received, not PT sold (the old integration test even documented "pass ptSold manually" as the workaround). First-ever cross-hook custom result interface in the codebase (verified: no precedent among ~60 prevHook call sites) — justified here, and the immutable pinning + operation/market/PT-identity checks make it safe.

Security posture: solid. Pin-before-external-call; market committed via inspect() (merkle) and cross-checked against the executed trade's market; PT identity re-derived from the committed market, not attacker-suppliable; NONE-operation (empty/stale context) rejected; cross-account isolation by account-keyed storage; executor mechanics confirmed (context bump per-hook at _processHook, so record.build reads PendlePTHook's still-current context; between the trade's postExecute and the oracle call nothing can move PT).


# Findings

## P1

### P1.1 aka F1 (merge-blocking, operational): PendlePTHook bytecode changes → full redeploy cascade 

The manifest currently codifies the old addresses for the new source. locked-bytecode/PendlePTHook.json is regenerated, but manifests/hooks.json still lists prod 0xb37CBfCd… (8 chains) / staging 0xCDBaC0d3… (5 chains) — deployments that do NOT have getPendleTradeResult. This is the identical failure class as PR 971 round 2 (Sonic/HyperEVM split-brain; that redeploy took two attempts). Required before/at merge, matching the ticket's own migration plan: redeploy PendlePTHook on all 13 chains, deploy both record hooks (bound to the new address), regen manifest AFTER, verify bytecode per chain, orphan hygiene (0xb37C joins 0x265f — never registry/SuperGovernor-whitelisted), merkle re-propose. Permanent coupling note: any future PendlePTHook redeploy now requires redeploying both record hooks too (immutable binding).


### P2.1 aka F2: CI is red, mostly inherited, but the sizing manifest looks hand-edited. 

test fail = AaveV3ArbitrumHooksFork setUp (dev-inherited from #978; dev CI red since). lint-manifest fail = regenerated hook-sizing-manifest.json contains AaveV3 entries missing from the committed file — also dev-inherited, BUT the PR's own 2 entries were evidently added by hand (a real regen would have pulled in the AaveV3 entries). All three new Pendle suites pass (31+22+20). deployment-smoke-test skipped while lint is red → the DeployV2Core changes haven't been smoke-validated yet. Fix: run the actual generator (also heals dev's staleness) or rebase after dev is fixed.


### P2.2 aka F3 (test gap vs ticket acceptance criteria): no E2E against the real PendlePTAmortizedOracleV2. 

The DETH-market fork test executes real router trades and asserts TradeResult == actual deltas (good), but the record hook's oracle call is only byte-compared against makeAddr("oracleV2") — never executed. Untested as a consequence: "oracle records the actual PT received" against real book-value math (incl. BOOK_VALUE_EXCEEDS_FACE_VALUE boundary), and the "TWAP minimum" criterion (twapDuration < min → TWAP_DURATION_TOO_SHORT) which no unit test covers either. Cheap fix: deploy a real V2 oracle in the fork test, execute the record execution as the account, assert bookValues/getTVLByOwnerOfShares.


### P3.1 aka F4: the two new manifest entries are internally inconsistent (pre-deploy generation artifacts). 

amountMeta = [{IN, TOKEN}] though the contracts' amountRoles() return empty (V2 entries correctly show []); erc165 = [] though the hooks support ISuperHookInflowOutflow (V2 entries list it). Presumably because the generator had no deployed contract to query. Must be clean after the post-deploy regen or erebor import may mis-derive params.



### P3.2 aka F5: spec/ticket vs implementation divergence on storage design. 

Ticket + specs say TradeResult lives in the existing nonce-keyed execution context ("offsets ≥5"); implementation uses a separate account-keyed transient namespace that is NOT nonce-scoped, so operation/market/input persist across context bumps within one tx while the output side resets. Unreachable in the executor flow (adjacency pin) and the purchase path fails closed (post-bump outputAmount reads 0 → AMOUNT_NOT_VALID), but either align the impl or update spec/natspec to document the deviation deliberately.



### P3.3 aka F6 (naming/registry): names drop "AmortizedOracle" (RecordPurchasePendlePTHook) 

This is clearer, but ticket says retain old hooks for historical decoding while preventing new strategies from selecting them; needs the registry-side deprecation flag. Solidity decoders are decodePtAmount/decodePtSold; erebor hook_params should still be named amount per Ronny (cf. the acknowledgePenalty naming saga).



# Downstream (not this PR related but still needed)

## D1

### D1.1 - From the previous I5 debug card 

The erebor registry entries for the two new hooks + bundling chain (PendlePTHook root → RecordPurchase next_required / RecordRedemption as same-seq alternate); NONE→NONE PASSTHROUGH hooks must not be offered as lane roots (guaranteed I5 400); OMS builder support (auto mode: usePrevHookAmount=true, amount=0, twapDuration default 900); merkle config re-propose with the NEW addresses. 

Ticket explicitly requires "register the new build paths and implementation addresses atomically with OMS support."

NicolaBernini (approved):
# Overview 
Overall LGTM but there are quire some findings that even if technically not blocking, would be better to tackle sooner than later, including the downstream one, see details here 
https://github.com/superform-xyz/v2-core/pull/982#issuecomment-5264495589

NicolaBernini:
# Overview

OK looks like a nice improvement 


# Details 

Round 1 recorded into PendlePTAmortizedOracleV2. 

Decision (Cosmin/Ronny/Sish): retarget to V1, the oracle prod actually reads for PT yield sources (mainnet 0xD64089698f82cbCD91ba5e0422aDFa81D247eB62; recording into V2 would update an oracle nobody consumes). 
V1's recordPurchase(market, sySpent, ptAmount) needs the cost in SY accounting-asset units, the exact parameter whose off-chain requirement caused the original OMS failure.

Solution: sySpent is computed on-chain as the oracle's own TWAP valuation 

```
sySpent = ORACLE.getAssetOutput(market, address(0), ptAmount) 
```

(V1's getPtToAssetRate(900) path; second param verified ignored). 

Properties, all verified:

P1) Self-consistent: the book increments by exactly how the oracle itself values the PT — no unit drift between recording and reads.

P2) Input-token-agnostic: buy route can pay in any token (fork test buys with mAPOLLO itself and asserts the valuation is unaffected by the pay token). TradeResult.inputAmount would have been in pay-token units — wrong for a book in accounting-asset units.

P3) Kills the off-chain-quote vector — the same class as the mAPOLLO/Kyber poisoned-quote incident ([[mAPOLLO PT round-trip toxic-pool loss]]).

P4) Cost-basis-vs-actual-cost tradeoff is design-accepted and bounded: rate ≤ 1e18 pre-maturity ⇒ sySpent ≤ ptAmount; face-value cap; per-msg.sender isolation. The committed security report proves same-block TWAP non-manipulability (Pendle writes the pre-swap lastLnImpliedRate once per block; OracleLib.write early-returns on same-block swaps) — only slow multi-block nudging of the actor's own fee baseline remains.


# Findings 

## From the previous round 

### F1 (redeploy cascade / manifest split-brain) → Operationalized. 

New committed specs/security-reports/2026-08-12-…-migration-runbook.md with the exact sequence (redeploy PendlePTHook on 8 prod + 5 staging → regen env output → deploy record hooks → regen manifests/hooks.json AFTER → per-chain bytecode verify → orphan hygiene 0xb37C/0xCDBaC/0x265f → merkle/registry re-propose), the downstream erebor/OMS list, and the permanent-coupling warning. Deploy script hardened: _resolveApprovedPendlePTHook computes the CREATE2 from THIS run's env-specific initcode, cross-checks script/output/{env} and warns on stale records, never binds to them. Still to execute — merge-gating in the operational sense only.


### F2 (CI / hand-edited sizing manifest) → Fixed. 

Real regen: 134 entries incl. the 7 missing AaveV3 (Cosmin added the AaveV3 overrides with researched offsets — SW 112 / BR 113 — fixing dev's inherited #978 lint break inside this PR). lint-manifest now GREEN both runs. Remaining test red = 1/5728: AaveV3ArbitrumHooksFork setUp, dev-inherited, unrelated.


### F3 (no real-oracle E2E; TWAP-min untested) → Fixed comprehensively. 

New PendlePTHookV1AmortizedOracleE2E (606 lines, 14/14 green in CI): mainnet fork on the mAPOLLO market, fresh V1 oracle AND the live prod 0xD64089 oracle; real buy→record asserting book == TWAP valuation ≈ actual USDC spent (±1%) and the IYieldSourceOracle surface; TWAP-below-min revert; multi-purchase accumulation over amortization; pull-to-par at maturity; full buy→sell→redeem lifecycle; multi-strategy + context isolation; no-position revert. Two honest fork workarounds, both documented: TWAP cardinality primed in setUp (see P3-3 below) and _freezeSyRate (Midas feed goes stale under multi-day warp).


### F4 (manifest amountMeta/erc165 inconsistency) → open by design; 

explicitly deferred to runbook step 4 (post-deploy regen). Fine.


### F5 (spec-vs-impl storage divergence) → Fixed in the right direction. 

The `_tradeKey` now includes BaseHook's per-account execution-context nonce (read from the same transient slot); TradeResult gets identical isolation to outAmount/outToken; natspec updated; context-isolation tests added (unit + both E2Es). New small fragility: ACCOUNT_CONTEXT_STORAGE = keccak256("hook.account.context") is duplicated from BaseHook (private there) with a MUST-match comment — verified matching today, but it's a silent-desync risk; suggest making BaseHook's constant internal (or an internal getter) for compile-time coupling.


### F6 (naming/deprecation) → carried into the runbook downstream list (registry deprecation flag for legacy hooks).


## New Findings 

All small 

### F7 No __checkContract entries for the two new hooks in the post-deploy verification section 

The legacy V1/V2 record hooks each have one (DeployV2Core ~L2100-2199); the smoke test therefore won't verify the new deployments. Add for parity (args = abi.encode(v1Oracle, approvedPendlePTHook) via the resolver).


### F8 Stale call-site comment 

The deploy block says "prefer the environment-recorded address … falling back to CREATE2" but _resolveApprovedPendlePTHook always binds computed (recorded is warn-only — the function natspec is correct). One line.


### F9 Runbook omission 

The downstream list has the decimals onboarding constraint but not the TWAP-warmth precheck (security report P3-3): at market onboarding run Pendle getOracleState(market, 900) → increaseObservationsCardinalityNext if needed — the fork suite proves the failure mode is real (mAPOLLO reverted OracleTargetTooOld until primed in setUp), and an unwarmed market DoSes the whole userOp including the already-executed swap.


### F10 Residual, documented, accepted 

V1's unit invariant is guarded at the purchase hook, not fixed in the oracle — markets with yield-bearing accounting semantics (getPtToAssetRate > 1e18) would still fail the face cap at first record (fail-closed); mismatched-decimals positions recorded by the LEGACY hooks remain possible on the same oracle. Both covered by the onboarding constraint.

PR #983, opened 2026-08-17
feat: euler hooks, morpho updates and aave natspec fixes

supervaulter:


src/hooks/loan/morpho/MorphoRepayAndWithdrawHookV2.sol
```diff
@@ -0,0 +1,376 @@
...
+
+        if (vars.isFullRepayment) {
+            // Full repay: share-based repay with post-accrual debt
+            (, uint128 borrowShares,) = morphoStaticTyping.position(ctx.id, account);
+            ctx.shareBalance = uint256(borrowShares);
+            if (ctx.shareBalance == 0) revert AMOUNT_NOT_VALID();
+
+            // Derive loan amount for approval (rounds up to cover accrued interest)
+            Market memory market = IMorpho(morpho).market(ctx.id);
+            ctx.approvalAmount = ctx.shareBalance.toAssetsUp(market.totalBorrowAssets, market.totalBorrowShares);
```
vikramarun: **Full repayment still has the pre-accrual/post-accrual mismatch the spec ordered fixed (high).**

The approval here is sized in `build()` from *stored* market totals, but `accrueInterest()` only runs afterward in `_preExecute` (line 366) — the executor's order is `build()` → `preExecute` → executions. So whenever the market has pending un-accrued interest at execution time (the normal case in production), the share-based `repay(0, shares)` pulls `toAssetsUp(shares, postAccrualTotals)` assets, which is strictly more than the allowance computed from stale totals → `transferFrom` reverts and the whole bundle fails.

The carry spec §9 lists exactly this — "full repayment can mismatch pre-accrual approval with post-accrual execution" — as one of the reasons the V1 composites cannot launch, and §9.2 requires "accrue the market **before** sizing approval or converting borrow shares." As written, V2 reproduces the flaw rather than fixing it (acknowledged in the NatSpec as KNOWN LIMITATION P1-3). Since full exit is a launch gate, this blocks the carry launch path.

The fork tests mask it: every V2 full-repay test either executes in the same block as open (no pending interest) or calls `accrueInterest` in a *prior* transaction (e.g. `test_fork_InterestAccrual_FullRepayIsExact` warps then accrues before executing). A test that warps **without** pre-accruing and then executes full repay should reproduce the revert.

Suggested fix: size the approval from accrual-adjusted expected totals (Morpho `expectedMarketBalances`-style preview, rounding up), or approve a small buffer above the stale estimate — mirroring how the Euler hook solved the same class of problem with `type(uint256).max`.

src/hooks/loan/morpho/MorphoRepayAndWithdrawHookV2.sol
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    /// @inheritdoc ISuperHookInspector
+    function inspect(bytes calldata data) external pure override returns (bytes memory) {
+        RepayWithdrawVars memory vars = _decodeV2Data(data);
+
+        MarketParams memory marketParams =
+            _generateMarketParams(vars.loanToken, vars.collateralToken, vars.oracle, vars.irm, vars.lltv);
+
+        return abi.encodePacked(
```
vikramarun, line 348: **Inspector still omits LLTV from the market identity (high).**

`inspect()` returns only `loanToken, collateralToken, oracle, irm`. The carry spec §9 explicitly names "the inspector omits LLTV from the complete market identity" as a launch blocker for the V1 composites, and requires the V2 hooks to carry the five canonical `MarketParams` fields as the complete identity. Two Morpho markets that differ only in LLTV are distinct markets (different market ID, different liquidation behavior) but are indistinguishable to Erebor leaf authorization under this inspect output.

Same issue in `MorphoSupplyAndBorrowHookV2.inspect()`. Both should include `lltv` (and the derived market ID would be even tighter) in the packed output.

src/hooks/loan/morpho/MorphoSupplyAndBorrowHookV2.sol
```diff
@@ -0,0 +1,220 @@
...
+        override
+        returns (Execution[] memory executions)
+    {
+        SupplyBorrowVars memory vars = _decodeV2Data(data);
+
+        if (vars.usePrevHookAmount) {
+            vars.primaryAmount = ISuperHookResult(prevHook).getOutAmount(account);
+        }
+
+        if (vars.primaryAmount == 0 || vars.secondaryAmount == 0) revert AMOUNT_NOT_VALID();
```
vikramarun, line 125: **No risk/cap bounds on the Morpho legs (medium).**

The Euler composite carries `maxPostDebt`, `maxLiqCapUtilBps`, and post-execution `accountLiquidity` checks; the repay side carries `maxRepayAssets` / `maxCollateralReleaseAssets` / a remaining-utilization bound. The Morpho V2 layouts have none of these — no post-borrow debt ceiling, no LLTV/liquidation-capacity utilization check, no inspected repay/release maxima, and no post-release health verification in `MorphoRepayAndWithdrawHookV2`.

Spec §9.1 requires the opening hook to "enforce inspected debt, LLTV/liquidation-capacity, market cash/cap, oracle, IRM, and slippage bounds," and §9.2 requires capping spend by an "inspected maximum" and verifying "debt reduction, collateral delta, and post-release health." Morpho is a day-one provider held to "the same implementation … gates as Euler," so the V2 hooks need tail fields and post-execute checks equivalent to the Euler ones.

src/hooks/loan/morpho/MorphoSupplyAndBorrowHookV2.sol
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    /// @inheritdoc BaseHook
+    function _preExecute(address, address account, bytes calldata data) internal override {
+        _setOutAmount(getCollateralTokenBalance(account, data), account);
+    }
+
+    /// @inheritdoc BaseHook
+    function _postExecute(address, address account, bytes calldata data) internal override {
+        _setOutAmount(getOutAmount(account) - getCollateralTokenBalance(account, data), account);
```
vikramarun: **Open hook returns collateral consumed instead of borrowed USDC (medium).**

`_postExecute` sets outAmount = collateral consumed and outToken = `collateralToken`. Spec §9.1's last requirement is to "return canonical USDC and the exact borrowed amount for aggregation with Euler output" — the Euler open hook does this (outToken = debtAsset, outAmount = borrowed delta), so the two providers' outputs currently disagree in kind. Any downstream consumer chaining on the previous hook's output (or the OMS validating aggregate borrowed-USDC conservation, §10) would read stock collateral units from Morpho where it expects USDC.

Suggested: track the loan-token balance delta and set outToken = loanToken, matching `EulerDepositCollateralAndBorrowHook`.

vikramarun (commented):
Review of the Morpho V2 hooks against the SuperStocks carry technical spec (§9). The Euler side tracks the spec closely, but the four comments below cover places where the V2 hooks miss the corrections the spec called for — including two of the three named reasons the V1 composites were declared unable to launch pooled carry.

supervaulter:
Close in favor of 992 and 990

PR #985, opened 2026-08-20
feat: HyperCore hooks for Hyperliquid perps (chain 999)

vikramarun:
## What

Six contracts letting a Superform smart account act as a Hyperliquid HyperCore master from HyperEVM (chain 999). The account **is** the HyperCore actor: CoreWriter's actor is `msg.sender` of `sendRawAction`, and Superform hooks execute as the account.

This is the contracts leg of the perps plan of record (§09c) — the on-chain surface for agent enrolment, margin movement, withdrawal, builder-fee approval, and the EVM→Core deposit. Trading itself is agent-signed to the Hyperliquid API and needs no hook.

| # | Contract | Action | Purpose |
|---|---|---|---|
| 00 | `ICoreWriter`, `IHyperCoreDepositGateway` | — | vendor interfaces |
| 01 | `BaseHyperCoreWriterHook` | — | envelope encoder `0x01‖uint24‖abi.encode(args)`; the audit surface |
| 02 | `HyperCoreAddApiWalletHook` | 9 | agent enrolment (device trading key) |
| 03 | `HyperCoreUsdClassTransferHook` | 7 | spot ↔ perp margin |
| 04 | `HyperCoreSendAssetHook` | 13 | withdraw leg + containment exit |
| 05 | `HyperCoreApproveBuilderFeeHook` | 12 | revenue model |
| 06 | `ApproveAndHyperCoreDepositHook` | — | EVM→Core deposit via per-token gateway |

**Status: code complete, security-reviewed, deployed to staging on 999. Audit is the remaining gate for mainnet.**

Only these six are needed. All 17 documented CoreWriter actions were swept against the product; five are token-agnostic and the deposit hook is per-token, so adding a second token is a deployment plus config, not new contracts. Sweep table is in §09c of the plan of record.

## Design decisions worth reviewing

- **Typed leaves over a generic `(actionId, bytes)` encoder** — narrows the reachable HyperCore surface to exactly four actions; the action-id list lives in one place in the base.
- **S2 sizing on every CoreWriter leaf**: `ISuperHookInflowOutflow` decode-only with empty `amountRoles`, `ISuperHookOutflow` explicitly refused — `replaceCalldataAmounts` does not exist on these contracts, so the bundler cannot rewrite a fee rate, an `ntl`, or a token index. `decodeAmounts`/`supportsInterface`/`_pipeMode` are all non-virtual so no leaf can regress.
- **`PASSTHROUGH` + `NONACCOUNTING`**: CoreWriter actions produce nothing EVM-observable. Deliberate consequence: no SuperLedger fee capture on funding flows — builder fees are the revenue model. Note the mechanism is `hookType`, not `_pipeMode` — `_updateAccounting` branches on the former.
- **No `usePrevHookAmount` on 03/04**: `ntl`/`wei` are HyperCore units; the EVM offset varies per token (`evmExtraWeiDecimals`: USDT0 −2, USDE +10, UETH +9). Chaining would silently misscale. Backend converts; hooks validate bounds.
- **Item 04 fixes three of six args**: `subAccount = 0` (zero in all 1,945 sampled payloads) and both dex fields `= type(uint32).max` (unanimous across all 84 sampled sends to a token system address). It also enforces destination↔token: a `0x20…`-band destination must match the token index packed in its low bytes, since that pairing strands funds and CoreWriter cannot revert. HYPE (`0x2222…2222`) sits outside the band and is documented as an exception.
- **Item 06 pins `TOKEN`, `GATEWAY` and `DESTINATION_DEX`** as immutables. The gateway credits `msg.sender`, so there is no recipient parameter to abuse. `tokenInfo.evmContract` is the **gateway**, not the token; the real USDC (`0xb88339…630f`) was verified on mainnet as the sole contract emitting `Transfer` into it.
- **Two counterintuitive validations are deliberate** (commented at the call sites): empty `agentName` is the default unnamed API-wallet slot, and `maxFeeRate == 0` is how a builder approval is revoked. Please don't "fix" these in review.

## Why the tests look the way they do

CoreWriter is 544 bytes: a gas-burn loop, one `LOG2`, return. **No validation, no revert path** — a malformed payload executes successfully and silently does nothing. Byte-exact fixtures against real mainnet transactions are therefore not defence-in-depth, they are the *entire* defence (an `abi.encodeCall` reconstruction shares the hook's own ABI assumption and passes against a wrong tuple).

- 5 fixtures assert payload bytes literally against decoded mainnet transactions (actions 9, 7, 12, 13×2 — including a non-zero-token action-13 sample so the token field is distinguishable from padding)
- **Mutation-tested**: action id 7→8 fails 2 tests; dex sentinel `0xffffffff`→0 fails both send-asset fixtures; the deploy call site `PERP`→`SPOT` fails the wiring test and nothing else
- Deploy-path tests deploy from **locked bytecode with the exact constructor args the deploy script assembles** and assert every immutable, including `TOKEN != GATEWAY`
- ABIs pinned from 2,073 decoded CoreWriter logs over 30k blocks, cross-checked against Hyperliquid docs
- Plus a fork suite (`test/integration/hypercore/HyperCoreHooksFork.t.sol`)

**59 unit tests pass.**

## Corrections landed during review

Three substantive errors were caught on this PR and fixed. Recording them because each has a generalisable shape.

| Was | Now | Why it mattered |
|---|---|---|
| Builder fee cap `1000` labelled "0.1%" | **`100`** (`_PERPS`) / `1000` (`_SPOT`) | `maxFeeRate` is **decibps** — 10 = 1bp. `1000` is 1%, the *spot* max. As deployed it would have admitted perp approvals at 10× the protocol limit. Two mutually-consistent sources agreed on the number and neither stated the unit. |
| Gateway `uint32` called "undocumented", deposits said to land in **spot** | **`DESTINATION_DEX`**; `0` = perp dex 0, `type(uint32).max` = spot | It is action 13's `destinationDex`. All **1,712** gateway-emitted payloads carry `(subAccount 0, sourceDex MAX, destinationDex 0)` — the evidence was already in our own sampling and the inference from it was wrong. Documented by [Hyperliquid](https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/hyperevm/interacting-with-hypercore) and corroborated by [Circle](https://developers.circle.com/cctp/howtos/transfer-usdc-from-arbitrum-to-hypercore). |
| `depositFor` said not to exist | Declared in the interface | `depositFor(address,uint256,uint32)` (`0xc23c545a`) is present and permissionless. The hook still uses `deposit()` deliberately — crediting `msg.sender` leaves no recipient parameter to abuse — but the capability is now on the record. |

**Consequence worth noting:** with `destinationDex = 0` the USDC deposit credits perp margin in one call, so the funding chain loses its `usdClassTransfer` leg and first deposits lose a HyperCore block. Hook 03 is still needed for withdrawals (perp→spot before the send) and for margin-defense tier 1.

## Token/dex pairing guard

A token paired with a dex it isn't quoted in emits an action 13 HyperCore won't credit — green receipt, stranded funds. The hook can't check this (it knows the token's EVM address, not its HyperCore index), so it's asserted in the deploy script.

Only **dex 0 and spot** are accepted. Dex 0's quote asset is known to be USDC so that pairing is enforced; every other perp dex is **rejected rather than inferred**, because under [HIP-3](https://hyperliquid.gitbook.io/hyperliquid-docs/hyperliquid-improvement-proposals-hips/hip-3-builder-deployed-perpetuals) a builder-deployed dex picks its own collateral — "not spot" does not imply "quoted in USDC". Supporting a HIP-3 dex is a decision someone must come back and make.

Both the guard and the **wiring** are tested: `_hyperCoreDepositArgsForChain` is asserted directly, so flipping the call site can't pass silently. Note the guard is convention, not enforcement — nothing stops a future call site encoding args inline instead.

## ⚠️ Staging: one address is stale

Staging on 999 (`script/output/staging/999/HyperEVM-latest.json`) was deployed at `0c84d252`. Comparing locked bytecode then vs now:

| Contract | Staging address |
|---|---|
| `HyperCoreAddApiWalletHook` | ✅ still valid |
| `HyperCoreUsdClassTransferHook` | ✅ still valid |
| `HyperCoreSendAssetHook` | ✅ still valid |
| `HyperCoreApproveBuilderFeeHook` | ✅ still valid |
| **`ApproveAndHyperCoreDepositHook`** | ❌ **stale** — `0x6cE04B6F…04b5` |

The `GATEWAY_ARG` → `DESTINATION_DEX` rename changes the public getter, hence the ABI, hence the bytecode, hence the CREATE2 address. **Redeploy that one.** Its staging instance was also built with the old `GATEWAY_ARG` semantics.

## Testing

```
forge test --match-path "test/unit/hooks/hypercore/*" --skip "test/integration/*"
```

The `--skip` is pre-existing on `dev` and unrelated: `MinimalBaseNexusIntegrationTest.t.sol` and `StargateAdapterE2EFork.t.sol` call `vm.getRecordedLogs()` from `view` helpers, which newer forge-std rejects — this blocks any bare `forge test` wholesale.

## Follow-ups (not in this PR)

- **Audit** before mainnet deployment on 999
- Redeploy `ApproveAndHyperCoreDepositHook` on staging
- Backend: `chains.yaml` + `hooks_config.json` entries, new `ActionType`s, and asserting **no `APPROVE` is prepended** (CoreWriter hooks touch no ERC-20; item 06 approves internally). Read `DESTINATION_DEX()` off the registered hook rather than assuming it.
- The two `view`/`getRecordedLogs` fixes on `dev`
- Builder must hold ≥100 USDC in its perp account to receive fees — Phase 3 setup, unowned


vikramarun:
## 🔒 Security review (automated, focused on this PR's changes)

**Result: no high-confidence vulnerabilities found.** Every attack surface specific to this change was verified; summary below. The one cosmetic issue found (brace indentation in `ConfigOtherHooks.sol`) is fixed in 4f2af315 along with `forge fmt`, with locked bytecode re-verified byte-identical after formatting.

### Verified

| Attack surface | Result |
|---|---|
| Byte offsets vs natspec layouts, all 5 hooks | ✅ Every offset hand-summed against field widths and strict `DATA_LENGTH` checks; no overlaps or off-by-ones; `BytesLib` bounds-checks fail closed |
| CoreWriter action ABI tuples (7, 9, 12, 13) | ✅ Match mainnet-transcribed fixtures and Hyperliquid docs, including action 9's dynamic-string encoding and action 13's pinned `subAccount`/dex fields |
| Envelope endianness (`bytes3(uint24)`) | ✅ Big-endian, confirmed by fixtures |
| Deposit hook approval lifecycle | ✅ `approve(0) → approve(amount) → deposit → approve(0)`, no standing allowance; `pre − post` delta is fail-closed (anomalous balance increase underflows and reverts); `NONACCOUNTING` so the ledger never consumes it |
| S2 sizing enforcement | ✅ `replaceCalldataAmounts` does not exist in any CoreWriter leaf's bytecode — the bundler cannot rewrite a fee rate, `ntl`, or token index; `_pipeMode()` non-virtual |
| Deploy-script constructor args | ✅ Order matches constructor signatures and locked-artifact ABIs |
| **Locked bytecode integrity** | ✅ **All 5 artifacts recompiled from source (solc 0.8.30, optimizer 200, prague) and byte-compared — exact matches** |
| `register_hypercore_hooks.sh` | ✅ Input is repo-owned deploy output, quoted, `cast`-validated; refuses to run unless the RPC reports chain 999 |
| Executor interplay | ✅ `validateHookCompliance` conformant; all executions `value: 0`; transient-storage mutexes intact |

### Documented residual risks (not vulnerabilities — reviewer attention wanted)

1. **`agentName` is an eviction/takeover key** (`HyperCoreAddApiWalletHook`). Re-enrolling an existing name replaces the prior agent key in ~6s. `inspect()` surfaces the agent *address* but cannot surface the name (inspectors are addresses-only by protocol rule). Both fields are committed by the user's signed Merkle root, and the natspec carries an explicit `SECURITY:` warning — **the OMS/backend must never allow this field to be attacker-chosen.** This obligation transfers to the superbundler integration.
2. **`HyperCoreSendAssetHook` cannot enforce destination↔token consistency** — sending token Y to token X's system address strands funds. Both fields are signature-committed and the destination is surfaced via `inspect()`; no on-chain check can distinguish system addresses without an oracle. Off-chain encoding responsibility.
3. **Hardcoded mainnet constants** (USDC `0xb88339…630f`, gateway `0x6B9E…0A24`, undocumented gateway `uint32 = 0`) should be independently re-verified at deployment time. The deployment test guards the token/gateway transposition; selector `0x2b2dfd2c` verified against `deposit(uint256,uint32)`.

### Two counterintuitive validations that are deliberate — do not "fix" in review

- `agentName` of **length zero is allowed**: it is Hyperliquid's default unnamed API-wallet slot.
- `maxFeeRate == 0` is **allowed**: it is how a builder approval is revoked.

Both are commented at the call sites.

vikramarun:
**Builder fee cap looks 10× too permissive — `HYPERCORE_MAX_BUILDER_FEE_RATE = 1000`**

Flagging this from the backend side: we copied the constant into `pkg/hypercore` with the same comment ("1000 == 0.1% == the perp protocol maximum") and it doesn't survive a check against Hyperliquid's docs.

`maxFeeRate` for CoreWriter action 12 is in **decibps** (tenths of a basis point), stated three independent ways in the docs:

- [Interacting with HyperCore](https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/hyperevm/interacting-with-hypercore), action 12: *"maxFeeRate is in decibps. To approve a builder fee of 0.01% maxFreeRate should be 10"*
- [Builder codes](https://hyperliquid.gitbook.io/hyperliquid-docs/trading/builder-codes): *"`f` is the builder fee to charge in tenths of basis points. I.e. a value of 10 means 1 basis point"*, and *"Builder fees charged can be at most 0.1% on perps and 1% on spot"*
- Info endpoint `maxBuilderFee` returns the approval in the same units (`1` = 0.001%)

So:

| decibps | percent |
|---|---|
| 10 | 0.01% (the docs' worked example) |
| **100** | **0.1% — perp maximum** |
| **1000** | **1% — spot maximum** |

`MAX_BUILDER_FEE_RATE` is an immutable constructor arg on `HyperCoreApproveBuilderFeeHook` and the only bound on the value the hook forwards, so as deployed it would admit perp builder-fee approvals up to 1%. Since CoreWriter never reverts, an over-cap approval is silently dropped on HyperCore behind a successful EVM receipt — exactly the failure mode the PR's own testing section is designed around, and one the locked bytecode makes expensive to fix later.

Suggested: `HYPERCORE_MAX_BUILDER_FEE_RATE = 100` for the perps deployment (or split perp/spot constants if a spot instance is planned), and update the natspec to say decibps. Happy to be wrong if you've measured HyperCore rejecting >100 on perps regardless of the hook's cap — but the hook's stated intent and its value disagree either way.

Backend side is corrected in superform-xyz/superbundler#448 (`MaxBuilderFeeRatePerps = 100`, `MaxBuilderFeeRateSpot = 1000`), along with a CI assertion on the hooks_config draft.

vikramarun:
You're right, and thanks — this was a real bug. Fixed in 563ee53f.

**Confirmed independently** against [builder codes](https://hyperliquid.gitbook.io/hyperliquid-docs/trading/builder-codes): *"a value of 10 means 1 basis point"*, and *"Builder fees charged can be at most 0.1% on perps and 1% on spot"*. So 0.1% = 10bp = **100 decibps**, and 1000 is the spot maximum. The constant was 10× too permissive for a perps deployment exactly as you describe.

**Where it came from**, since it's worth recording: the plan of record's builder-fee decision says *"approve the cap at 1000 (0.1%, the protocol max)"*, and my mainnet harvest independently observed `maxFeeRate=1000` on live action-12 payloads. Those two agreed with each other, so I took the unit as settled — but the observation only shows a builder approving 1%, which is spot-legal. Two consistent sources, both silent on the unit. `maxFeeRate` units were on my own open-questions list earlier in this work and I wrote the constant without closing that item.

**Note the same 10× appears twice in the plan of record**, which suggests one unit misunderstanding rather than a typo — §10 also says *"charge 250 (0.025%)"*, but 250 decibps is 0.25%, which is **above the 0.1% perp cap**. Worth correcting there too; I've left that to whoever owns the revenue decision rather than editing the number myself.

## What changed

- Split into `HYPERCORE_MAX_BUILDER_FEE_RATE_PERPS = 100` and `_SPOT = 1000`, matching your `superbundler#448` naming so the two don't drift. Deploy passes the perps value.
- Natspec now states decibps with the conversion, and calls out the trap explicitly — including that the failure is silent because CoreWriter never reverts.
- **The mainnet action-12 fixture now builds against a spot-capped instance.** It carries `maxFeeRate=1000`, and that value being spot-legal rather than perp-legal is precisely why the perps cap is 100 — so the fixture now documents the distinction instead of obscuring it.
- **Two regression guards**: the perps cap must reject the spot maximum, and must accept 0.1% expressed correctly. I verified the first has teeth by reverting the constant to 1000 — it fails with `next call did not revert as expected`, then passes again once restored. It would have caught this.

25/25 tests pass.

## One correction in your favour

> the locked bytecode makes expensive to fix later

`MAX_BUILDER_FEE_RATE` is a **constructor immutable**, supplied by the deploy script from `ConstantsOtherHooks` — the value was never in the bytecode. All five locked artifacts verified byte-identical after this change (natspec doesn't affect bytecode under `bytecodeHash: none`, but I compared rather than assumed). So this was a one-constant fix, not a re-lock.

Worth generalising though: **a cap that lives in a constructor arg is cheap to correct; one baked into a constant inside a hook would not be.** That's an argument for keeping protocol limits as deploy-time immutables in this family, which is what the other hooks already do.

vikramarun:
**Correction to my previous comment** — I had the builder *charge* rate backwards.

I wrote that §10's *"charge 250 (0.025%)"* meant the percentage label was wrong and the real charge would be 0.25%. It's the other way round: **0.025% is the intended rate, so the integer is what's wrong.**

At the docs' own worked example — 0.01% → `10` — the conversion is:

| decibps | percent |
|---|---|
| 10 | 0.01% |
| **25** | **0.025% — the intended charge** |
| 100 | 0.1% — perp cap |
| 250 | 0.25% |
| 1000 | 1% — spot cap |

So the charge constant should be **25**, which sits comfortably under the 100 cap rather than 2.5× over it. Same 10× unit error as the cap, and it does confirm both numbers came from one misunderstanding rather than two typos.

**No contract change.** The charge rate is the per-order `builder` field, not a hook parameter — `HyperCoreApproveBuilderFeeHook` only enforces the *cap*, which is already fixed at 100 in 563ee53f. This is a backend/plan-of-record correction only, flagged here so the number doesn't get encoded as 250 anywhere.

supervaulter (approved):

vikramarun:
**`GATEWAY_ARG` is HyperCore's `destination_dex` — and `0` credits perp margin, not spot**

Backend side again. `IHyperCoreDepositGateway` calls the second argument "undocumented", notes that every observed mainnet call passes `0`, and correctly rules out it being a token index. That's all right, but the conclusion that it lands in spot isn't — `0` is perp dex 0.

**The answer is already in the sibling file.** `HyperCoreSendAssetHook` names `type(uint32).max` as `NO_DEX`, "the no perp dex sentinel". The gateway's `arg` is that same field: it forwards to CoreWriter action 13 and passes `arg` straight through as `destinationDex`.

Evidence:

1. The gateway calls CoreWriter itself — selector `0x17938e13` is present in its implementation (`0x7537af00779cc053a696e47ebd451b5bc4790da3`, behind the transparent proxy at `0x6b9e7731…0a24`).
2. I decoded every `RawAction` the gateway emitted in a 150-block window on mainnet — 21 of them, unanimous:
   ```
   action 13 (destination, subAccount=0, sourceDex=0xFFFFFFFF, destinationDex=0, token=0, wei)
   ```
   Gateway's own spot → the depositor's **perp** account.
3. Non-gateway action-13 traffic in the same window corroborates the field positions: `(…, 0xFFFFFFFF, 0xFFFFFFFF, …)` ×6 (spot→spot sends) and `(…, 0, 0, …)` ×3 (perp→perp). Only the gateway shows the mixed pair.
4. [Hyperliquid's action table](https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/hyperevm/interacting-with-hypercore) gives action 13 as `(address, address, uint32, uint32, uint64, uint64)` and "specify uint32::MAX for the source_dex or destination_dex for spot".
5. Circle documents the same knob for their HyperCore forwarder as `hyperCoreDestinationDex`: "by default (when `hyperCoreDestinationDex` is `0`), deposits credit the perps balance"; `4294967295` for spot. So the path isn't undocumented after all — [it's just documented by Circle](https://developers.circle.com/cctp/concepts/cctp-on-hypercore) rather than by Hyperliquid.

Our own mainnet spike passed `0xFFFFFFFF` and landed in spot, which is consistent with all of the above and is why we read it as a spot-only gateway in the first place.

**What this affects**

The good news is that `GATEWAY_ARG` being a constructor immutable means nothing needs to change in the logic — this is a deploy-config decision. But three doc strings currently point a deployer at the wrong value:

- `IHyperCoreDepositGateway` title: "credits an EVM sender's HyperCore spot balance"
- `ApproveAndHyperCoreDepositHook` title and `description()`: "Deposits an ERC-20 into a HyperCore spot balance"
- The `@param arg` / `GATEWAY_ARG` natspec: "undocumented", "should its meaning ever be documented"

Suggestion: rename to `DESTINATION_DEX`, and let the per-token deployment pattern already in the natspec carry the right value per token — `0` for the USDC perps-funding instance, `type(uint32).max` for spot tokens (PURR, HFUN, UBTC, MUX). That fits the "one instance per token, CREATE2 makes it near-free" design as-is.

Worth doing because deploying the USDC instance with `destinationDex = 0` **removes a whole leg from the funding chain**: the deposit credits perp margin in one call, so no spot landing and no follow-up `usdClassTransfer`. Given HyperCore won't let an account be funded and act in the same block, that's one fewer block on every first deposit.

**Separately — `depositFor` does exist**

`IHyperCoreDepositGateway` says "Credits msg.sender; there is no recipient-taking overload." There is one: `depositFor(address,uint256,uint32)`, selector `0xc23c545a`, present in the implementation. It's permissionless — an `eth_call` from an address holding nothing reverts with `ERC20: transfer amount exceeds allowance`, i.e. it reaches the token pull without hitting any auth check. The caller supplies the tokens, the named recipient gets credited.

Not asking for the hook to use it — crediting `msg.sender` is the better security property here and the "no address parameter to supply, and equally none to abuse" reasoning holds. Just flagging that the doc claim is wrong, and that this is what lets us pre-activate an account's HyperCore side from the backend without a UserOp.

**`HyperCoreSendAssetHook` — no ask, one consequence**

That hook documents all six action-13 arguments and pins three with sampling evidence, which answered the question I came in with. Worth stating the downstream consequence explicitly though: with both dexes pinned to spot, perp margin can't be withdrawn in one hop — it needs `usdClassTransfer` perp→spot first, then the send. That's the documented path and I think the narrowing is the right call for v1; just noting it so nobody plans a one-hop perp withdrawal against this hook.

vikramarun:
Confirmed and fixed in `4811ed64` + `d431e137`. You're right on all three points, and the first one is the one that stings — **the evidence was already in my own sampling and I drew the wrong inference from it.**

## Verified independently

- **Gateway emits `destinationDex = 0`.** Re-checked my harvest: all **1,712** gateway-emitted action-13 payloads carry `(subAccount 0, sourceDex 0xFFFFFFFF, destinationDex 0)` — unanimous. I had run exactly this analysis earlier in the work, saw the mixed pair, and correctly read `0xFFFFFFFF` as the no-dex sentinel for the *withdraw* hook — while never connecting that the gateway's *mixed* pair meant deposits land in perp.
- **Hyperliquid docs**, action 13: *"Specify uint32::MAX for the source_dex or destination_dex for spot."* So `0` is perp dex 0. Decisive.
- **`depositFor(address,uint256,uint32)`** — selector `0xc23c545a` confirmed present in the implementation. My "there is no recipient-taking overload" was simply wrong.

(Circle's CCTP-on-HyperCore page doesn't actually document `hyperCoreDestinationDex` — I checked. The Hyperliquid quote and the 1,712 samples carry it without needing that.)

## Changed

- `GATEWAY_ARG` → **`DESTINATION_DEX`**; constants split into `HYPERCORE_DESTINATION_DEX_PERP` (0) and `_SPOT` (`type(uint32).max`), so the per-token deployment pattern carries the right value as you suggested
- Interface and hook no longer claim deposits land in spot; the `0` = perp / `MAX` = spot semantic is stated with the doc quote
- `depositFor` **declared** in the interface with a note that the hook deliberately doesn't use it — crediting `msg.sender` keeps the "no recipient parameter to abuse" property, and recording it means the backend pre-activation path is discoverable rather than lost
- **Re-locked `ApproveAndHyperCoreDepositHook`** — the public getter rename changes the ABI and therefore the bytecode. Other four verified byte-identical.

## Notes on the merge

I rebased onto your work rather than force-pushing. Your `HyperCoreSendAssetHook` destination↔token check is a genuinely better outcome than what I left: I'd written that pairing off in the security review as *"not enforceable on-chain without an oracle"*, and enforcing it within the `0x20…` band — with HYPE's out-of-band address called out — closes it. Same for making `decodeAmounts`/`supportsInterface` non-virtual; that was an inconsistency in my base.

I also carried the rename into the two newer test files (`ApproveAndHyperCoreDepositHook.t.sol`, `HyperCoreHooksFork.t.sol`) so nothing dangles. 52/52 pass.

**One thing worth a look given the staging deploy:** renaming a public immutable changes the ABI, so the CREATE2 address of `ApproveAndHyperCoreDepositHook` moves. Anything already deployed to staging under the old bytecode is now at a stale address.

**And the consequence you flagged is the valuable part** — a USDC instance at `destinationDex = 0` credits perp margin in one call, so the funding chain loses its `usdClassTransfer` leg entirely and first deposits lose a HyperCore block. `HyperCoreUsdClassTransferHook` is still needed for withdrawals (perp→spot before the send) and for margin-defense tier 1, so it isn't redundant — just off the deposit path. I'll reflect that in the plan of record's money path unless you'd rather own that edit.

vikramarun:
## Per-instance keying for the HyperCore deposit hook (`947729fc`)

Landing the convention for how hooks-resolution picks between per-token deposit hooks. Not needed for v1 — we're USDC-only — but it goes live the moment a second token is registered, and it's free to change now and not later.

### Why this isn't a selector gap

`ApproveAndHyperCoreDepositHook` pins `TOKEN`, `GATEWAY` and `DESTINATION_DEX` as immutables, so a second collateral token or a spot destination means a second deployment of the same contract. That makes it **the first hook in this library that isn't one-per-chain.** Every hook that varies per target takes the target as a data field — `ApproveAndDeposit4626VaultHook` has no immutables and serves any vault. The hooks that *do* pin immutables pin protocol singletons (the Odos router, CCTP contracts), one per chain. Across prod deployment outputs the only suffixed keys are version suffixes (`StargateSendHookV2`), never instance suffixes.

Two layers assume one-per-chain, and both break quietly rather than loudly:

1. **The deploy record can't hold two.** `__deployContract` writes `contractAddresses[chainId][contractName]` and exports keyed by name — a flat map. Two instances under one name overwrite each other in `HyperEVM-latest.json` before resolution ever sees them.
2. **Resolution has no token dimension.** `selectHook` sorts by `IsApprovalHook` then `PreferenceScore`; instances sharing an `ActionType`, chain and `isApprovalHook` tie on every field, so the winner is arbitrary.

The failure mode is the bad one. If the USDC instance is handed a USDT0 funding intent and the account holds no USDC, it reverts loudly — fine. If the account *also* holds USDC, which is likely for a perps user, it approves and deposits **USDC** in an amount computed for USDT0, and the bridged USDT0 sits on HyperEVM. Wrong asset moved, green receipt, no error. Same shape as everything else in this integration: CoreWriter can't revert, and neither can this.

### The change

The deployment key now names what the instance is pinned to:

| | before | after |
|---|---|---|
| Deployment key / CREATE2 salt | `ApproveAndHyperCoreDepositHook` | **`ApproveAndHyperCoreDepositUsdcPerpHook`** |
| Contract name / locked bytecode | `ApproveAndHyperCoreDepositHook` | unchanged |

The two being different is the point, not a typo — the key names an *instance*, the bytecode lookup names a *contract*. Both scripts that list these hooks were already keyed correctly for their own purpose and stay that way: `deploy_v2_other_hooks_staging_prod.sh` checks bytecode files so it keeps the contract name; `register_hypercore_hooks.sh` resolves addresses from deployment output so it takes the new key.

Adding a token from here is: a new key constant, a new deploy entry, a new `ActionType`. The key doubles as the CREATE2 salt, so distinct keys give distinct addresses for free.

### Why a key rather than a tag

The alternative was a per-token tag on the registry record matched at resolution time, which needs no proto change. I'd argue against it: **the hook pins an address and a tag matches a symbol.** That's two identities for one concept with the mapping living in a naming convention nothing verifies, and symbols aren't unique here — the HyperEVM USDC ERC-20, HyperCore spot USDC, and USDC on other chains are three different things wearing one label. It's the same failure class as the pairing guard earlier in this PR: a check keyed on something *adjacent to* what it protects rather than identical to it.

A `token` field on `GetHookRequest` matched against the `TOKEN` immutable is the most precise option and generalises best if a proto change happens for other reasons. But it solves a disambiguation problem that per-key registration prevents from existing.

### The key is (token, dex), not token

Confirmed from the deploy guard: it permits exactly **dex 0 with the quote asset** and **spot with any token**. So a USDC-into-spot instance can legitimately coexist with USDC-into-perp, while UBTC-into-perp is rejected at deploy time. Hence `UsdcPerp` rather than `Usdc` — naming both dimensions now, even though only one instance exists, is precisely the retrofit we're avoiding.

For v1 no spot instance should be deployed: `destinationDex = 0` is the whole point of the funding simplification, and v1 never funds a spot balance.

### Tests

Two, over the constants the deploy path actually reads rather than local copies:

- `test_HookKeysAreUnique` — pairwise distinct across all five HyperCore keys. The realistic failure when the next token is added is a copy-pasted key, and `__deployContract` overwrites on collision rather than reverting.
- `test_DepositHookKeyIsInstanceScopedNotContractScoped` — pins the key/contract-name split so it can't be "tidied" back into a collision.

Both mutation-checked (duplicated the key, confirmed both fail, restored). 66 tests pass.

### Two things for others

- **@backend** — the key becomes an `ActionType`, so the exact string is yours to accept or change; I'll rewire to whatever you prefer. Note `manifests/hooks.json` and `hook-classification.yaml` stay keyed by *contract* name, which I think is right — one entry describes the contract's shape, and the per-key registration is what creates distinct records. Worth confirming that's how you want to consume it.
- **@protocol** — I left the staging record under the old key untouched: `0x6cE04B6F…04b5` is a real deployment and that file is script-generated, so hand-editing it would be falsifying the record. It's already stale from the `destinationDex` rename, and its replacement lands at a new address under the new key, so it wants cleaning up on the next staging deploy rather than now.

vikramarun:
## Proposed, not landed: derive the resolution tag from the hook key

Backend flagged the remaining gap on per-token deposit hooks ([context](https://github.com/superform-xyz/v2-core/pull/985#issuecomment-5359244144)):

> `test_HookKeysAreUnique` guards collision at deploy; nothing guards ambiguity at resolution, and that one bites at runtime rather than at deploy time.

They've added a `usdc` tag to their config entry as the selection key, and asked contracts to own the convention so the tag goes on with the second instance. Posting the proposed script change here rather than committing it, so the convention gets agreed before it ships.

### The idea

The key already encodes both dimensions — `ApproveAndHyperCoreDepositUsdcPerpHook`. So the tag shouldn't be a step anyone has to *remember* alongside registration; it should be **read off the key**. Then it can't be forgotten, and it can't disagree with what was actually deployed.

`register_hypercore_hooks.sh` parses `<Token><Dex>Hook`, refuses to register a deposit instance whose key doesn't match, and prints a copy-pasteable JSON block at the end.

### One thing worth deciding before this lands

**A token-only tag stops working the moment a second USDC instance exists.** `…UsdcPerpHook` and `…UsdcSpotHook` would both tag `usdc` and tie again — and that pairing is explicitly supported: the deploy guard permits dex 0 with the quote asset *and* spot with any token.

So the proposal emits **two** tags, `usdc` and `perp`. If resolution selects on a single tag, that either wants to be one composite (`usdc-perp`) or the selector needs both. That's a resolution-side call — happy to emit whatever shape you want, the derivation is the same either way.

### The diff

```diff
diff --git a/script/run/others/register_hypercore_hooks.sh b/script/run/others/register_hypercore_hooks.sh
index a7e3d570..34f4ec71 100755
--- a/script/run/others/register_hypercore_hooks.sh
+++ b/script/run/others/register_hypercore_hooks.sh
@@ -70,6 +70,54 @@ if [ "$missing" -ne 0 ]; then
     exit 1
 fi
 
+# ── Derive selection tags from the hook keys ───────────────────────────────────
+# ApproveAndHyperCoreDeposit instances are keyed <Token><Dex>Hook because one
+# contract ships once per (token, destinationDex) pair. Hook resolution has no
+# token dimension of its own and selects on tags instead — and a tag that has to
+# be REMEMBERED when a second instance is registered is the weakest possible
+# guard for a failure that only appears at runtime. So derive it from the key:
+# it then cannot be forgotten, and cannot disagree with what was deployed.
+#
+# Both dimensions are emitted deliberately. A token-only tag stops disambiguating
+# the moment a spot instance ships beside the perp one, which is a supported
+# combination — the deploy guard permits dex 0 with the quote asset, and spot
+# with any token.
+DEPOSIT_KEY_PREFIX="ApproveAndHyperCoreDeposit"
+
+declare -A HOOK_TAGS
+for name in "${HOOK_NAMES[@]}"; do
+    case "$name" in
+        "$DEPOSIT_KEY_PREFIX"*)
+            if ! tags=$(python3 - "$name" <<'PY'
+import sys
+key = sys.argv[1]
+PREFIX, SUFFIX = "ApproveAndHyperCoreDeposit", "Hook"
+DESTINATIONS = {"Perp": "perp", "Spot": "spot"}
+if not key.endswith(SUFFIX):
+    sys.exit("key does not end in '%s'" % SUFFIX)
+middle = key[len(PREFIX):-len(SUFFIX)]
+for word, dex in DESTINATIONS.items():
+    if middle.endswith(word):
+        token = middle[: -len(word)]
+        if not token:
+            sys.exit("no token segment before the destination")
+        print("%s %s" % (token.lower(), dex))
+        break
+else:
+    sys.exit("destination must be one of %s" % sorted(DESTINATIONS))
+PY
+            ); then
+                echo "❌ $name does not match ${DEPOSIT_KEY_PREFIX}<Token><Dex>Hook"
+                echo "   The key is the only place this instance's token and destination are"
+                echo "   recorded, so resolution cannot tell it apart from another instance"
+                echo "   without them. Rename the key and redeploy rather than tagging by hand."
+                exit 1
+            fi
+            HOOK_TAGS[$name]="$tags"
+            ;;
+    esac
+done
+
 # ── Load RPC ───────────────────────────────────────────────────────────────────
 if [ -z "${HYPEREVM_MAINNET:-}" ]; then
     echo "Loading HyperEVM RPC from 1Password..."
@@ -178,4 +226,30 @@ echo "  Skipped (existing):   $skipped"
 echo "  Failed:               $failed"
 echo "============================================================"
 
+# ── Emit derived selection tags for the hooks-resolution config ───────────────
+if [ "${#HOOK_TAGS[@]}" -gt 0 ]; then
+    echo ""
+    echo "Selection tags, derived from the keys — copy into the hooks-resolution config."
+    echo "These are not a naming suggestion: an instance whose config entry omits them is"
+    echo "indistinguishable from any other deposit instance at selection time, and the"
+    echo "wrong pick moves the wrong asset without reverting."
+    echo ""
+    {
+        echo "{"
+        n=0
+        total=${#HOOK_TAGS[@]}
+        for name in "${HOOK_NAMES[@]}"; do
+            [ -n "${HOOK_TAGS[$name]:-}" ] || continue
+            n=$((n + 1))
+            read -r token dex <<< "${HOOK_TAGS[$name]}"
+            comma=","
+            [ "$n" -eq "$total" ] && comma=""
+            printf '  "%s": { "address": "%s", "tags": ["%s", "%s"] }%s\n' \
+                "$name" "${HOOKS[$name]}" "$token" "$dex" "$comma"
+        done
+        echo "}"
+    }
+    echo ""
+fi
+
 [ "$failed" -eq 0 ]
```

### Sample output

```
Selection tags, derived from the keys — copy into the hooks-resolution config.
These are not a naming suggestion: an instance whose config entry omits them is
indistinguishable from any other deposit instance at selection time, and the
wrong pick moves the wrong asset without reverting.

{
  "ApproveAndHyperCoreDepositUsdcPerpHook": { "address": "0x…", "tags": ["usdc", "perp"] }
}
```

And the failure path, if someone adds a UBTC instance as `ApproveAndHyperCoreDepositUbtcHook` (no destination segment):

```
❌ ApproveAndHyperCoreDepositUbtcHook does not match ApproveAndHyperCoreDeposit<Token><Dex>Hook
   The key is the only place this instance's token and destination are
   recorded, so resolution cannot tell it apart from another instance
   without them. Rename the key and redeploy rather than tagging by hand.
```

### Open questions

1. **One tag or two?** Above.
2. **Should the script write a file** rather than print, so config generation consumes it directly instead of someone copying? Prints today because the deploy pipeline has no such artifact and I didn't want to invent one unilaterally.
3. **Is a Solidity-side key-shape test wanted too?** I'd proposed one asserting any `ApproveAndHyperCoreDeposit*` key parses, so the failure lands at `forge test` rather than at registration time. Not in this diff — it only pays off if the shell check isn't already the gate.

No rush on any of it: with one instance registered nothing is ambiguous today.

vikramarun:
## Stale staging key removed — no need to file the ask (`69d4f2ff`)

Done rather than queued: `ApproveAndHyperCoreDepositHook → 0x6cE04B6F…04b5` is out of `script/output/staging/999/HyperEVM-latest.json`. Staging now carries the same five HyperCore entries prod does.

**Why it survived the rename**, since it's the part worth remembering: the deposit hook changed **key**, not just bytecode. `SendAsset` kept its key, so its record was overwritten in place when it moved. The deposit hook's old entry had nothing to overwrite it, so it persisted as an orphan alongside the new one. Any future per-instance rename has the same shape.

**On the classification argument** — agreed, and it's the right reason to remove it rather than leave it. Unclassified is fail-safe for resolution, which can never select it, but `containsPerpsHook` and `fundsAndActsInOneBlock` key off `ActionType`, so a leg using that address directly would be invisible to the perps gate and the activation check. That's a narrower version of what `perps_gate.go:17` already documents, but this one costs a single line to close.

**Correcting myself:** an earlier revision of the plan of record argued *against* exactly this, on the grounds that the file is a record of a real deployment and hand-editing it would falsify the record. That was too precious. This file is a **handoff artifact** for downstream consumers, not an audit log of chain history — the chain itself and this repo's git history are both intact records that `0x6cE04B6F…04b5` exists. And the deploy script no longer emits that key, so the deletion is stable rather than something the next deploy undoes.

### Two things I checked while in there

**Prod is deployed.** `27cfa8d4` put all five on prod as well, and they're live on chain 999 with correct immutables — prod deposit hook `0xCF90E02e…784D` reads `DESTINATION_DEX() = 0` and `TOKEN() = 0xb88339CB…630f`, prod builder-fee hook `0xe4cF6859…164e` reads `MAX_BUILDER_FEE_RATE() = 100`. Prod carries only the new key; it was never affected by the orphan.

**Nothing is registered yet.** None of the seven HyperCore addresses across both environments appears in SuperGovernor's registry (checked against the live 55-entry list). So the orphan was inert on-chain today regardless — the exposure was always config generation reading this file, which is exactly what the removal closes.

Worth flagging separately that "prod waits on the audit" is no longer accurate as a description of *deployment* — the contracts are on prod, unregistered. Whether registration should wait on the audit is a call for whoever owns that gate, not something I'd assume either way.

PR #988, opened 2026-08-28
feat: [SUP-20796] v2 versions

supervaulter:

PR #990, opened 2026-08-31
feat: [SUP-20797] Euler EVK collateral, borrow and repayment hooks

supervaulter:
> [!NOTE]
> **Stacked PR** — based on #988 ([SUP-20796] v2 versions) and must merge after it. This PR's base branch is `cosmin-sup-20796-feature-standardize-versioned-position-loan-hooks-and`; once #988 merges, retarget to `dev`.

Linear: [SUP-20797](https://linear.app/superform/issue/SUP-20797/feature-add-the-euler-evk-collateral-borrow-and-repayment-hook-suite)

## What

First Euler EVK/EVC loan hooks in v2-core (Base / Clearstar), built on the SUP-20796 standardized V2 loan-hook machinery:

- `src/hooks/loan/euler/BaseEulerLoanHook.sol` — shared 197-byte layout, strict decoding, pinned-singleton validation, state-derived release verification
- `src/hooks/loan/euler/EulerDepositCollateralAndBorrowHook.sol` — deposit collateral + enable collateral/controller on the EVC + borrow (exact both legs)
- `src/hooks/loan/euler/EulerRepayHook.sol` — standalone repay up to a cap (`min(cap, debtOf)`), vault-path controller disable on clear
- `src/hooks/loan/euler/EulerRepayAndWithdrawHook.sol` — repay then withdraw exact collateral assets, releasing controller/collateral flags on clear
- Vendor: minimal `IEVC`, `IEVault`, `IGenericFactory` interfaces (license-clean surfaces)
- Deploy: `runEulerHooks` / auto-deploy gated on per-chain Euler config (`ConfigOtherHooks.eulerEvcs`), constructor args per the Morpho pattern; generated + locked bytecode; hook manifests regenerated

## Security review (in-repo, pre-merge)

Full `/superform:security` review (3-agent) ran on these hooks — report at `specs/security-reports/2026-08-28-euler-loan-hooks.md`: **0 P0/P1, 3 P2, 12 P3**, all remediated in this PR except one accepted P3 (raw `IERC20.approve`, Morpho-consistent, fail-closed). Highlights:

- **Trust pinning:** the canonical per-chain EVC and eVaultFactory are constructor-pinned immutables; calldata `evc` must equal the pin (`EVC_NOT_CANONICAL`) and every configured vault must be a factory proxy (`UNTRUSTED_VAULT`). A hostile contract self-reporting the canonical EVC/asset can never be enabled as controller.
- **State-derived release verification:** no fail-open transient flags — `_postExecute` re-derives from live chain state (`debtOf == 0` ⟹ controller/collateral disabled), unpoisonable by interleaved executions.
- **Graceful zero debt:** a third party gifting a full repayment can no longer cancel a signed intent (repay leg skipped, zero settle).
- **Fail-fast doomed closes:** residual debt + full-collateral withdrawal reverts `RESIDUAL_DEBT_FULL_WITHDRAW` at resolution time instead of an opaque EVK health-check revert.
- **Donation-proof collateral release:** `disableCollateral` fires on any predicted clear while enabled (no fragile `previewWithdraw == balanceOf` equality); protects the EVC 10-collateral cap.
- **Honest ERC-165:** the standalone repay hook does not advertise `ISuperHookLoans` (its reserved-zero collateral slot makes the inherited collateral getter revert). `BaseLoanHook._supportsLoanInterface` became `virtual` — locked bytecode of deployed hooks verified byte-identical.

## Testing

- `test/unit/hooks/loan/EulerLoanHooks.t.sol` — 91 tests (mock EVC/EVault/factory; strict decode, bindings, pinning, cap semantics, settle round-trips, release checks)
- `test/unit/hooks/loan/EulerLoanHooksForkBranchCoverage.t.sol` — 48 tests against real Base state @ block 50,550,000
- `test/integration/euler/EulerLoanHooksFork.t.sol` — 15 E2E tests through the full UserOp flow (open/chained open/repay/close, accrual, revert matrix)
- Full `test/unit/hooks/loan/*`: 654 pass (Morpho/Aave suites unaffected)

https://claude.ai/code/session_01Go7tfsFtQ5c3EtKH9zZvCP


NicolaBernini:
# Review 1 - Overview 

The #988 is a good standardization effort 

This PR adds a pinned-EVC / pinned-eVault-factory Euler base plus three hooks:
- deposit collateral + borrow;
- repay;
- repay + withdraw.

The positive controls are substantial:
- the Base EVC and GenericFactory are constructor-pinned;
- every configured eVault must be a factory-recognized proxy;
- every vault must report the pinned EVC and expected underlying asset;
- the composite inspector commits to EVC, controller vault, collateral vault, debt asset and collateral asset;
- controller/collateral lifecycle cleanup is explicit after a predicted full close;
- repayment is capped to live debt and zero debt is handled gracefully.



# Findings 

## B1 is P0 — Euler full-withdraw guard incorrectly collapses a multi-collateral account to one collateral

Location: [EulerRepayAndWithdrawHook.sol, reviewed head, lines 183–200](https://github.com/superform-xyz/v2-core/blob/281af058c3b2763a9a21e940a89df7287e8a142d/src/hooks/loan/euler/EulerRepayAndWithdrawHook.sol#L183-L200)

The hook computes whether the controller debt will be fully cleared. If residual debt is predicted and previewWithdraw(withdrawAmount) >= balanceOf(account), it reverts with RESIDUAL_DEBT_FULL_WITHDRAW.

That premise is valid only for a one-collateral account. It is false under Euler's actual EVC model:
- an EVC account may enable up to ten unique collaterals;
- account-status checks pass the controller the complete current collateral array;
- therefore the account can safely withdraw 100% of collateral A while collateral B (and potentially C…J) keeps the account healthy.

Euler's own health check is the canonical arbiter. The Superform hook rejects before that check, so it blocks a valid, provider-supported operation. It also forces a dust workaround—leave at least one share in collateral A—even when the account is healthy without A.

Euler evidence: [IEthereumVaultConnector.sol — collateral set and account-status interface](https://github.com/euler-xyz/ethereum-vault-connector/blob/838e5f72eaea25fab7d242760245244226096054/src/interfaces/IEthereumVaultConnector.sol#L199-L221) and [status check receives the full collateral array](https://github.com/euler-xyz/ethereum-vault-connector/blob/838e5f72eaea25fab7d242760245244226096054/src/interfaces/IEthereumVaultConnector.sol#L395-L401).

Recommended minimum fix
B1.R1) Remove the single-collateral pre-revert and let the eVault/EVC withdrawal perform the canonical post-operation health check; or, if an early check is needed, base it on Euler-native maxWithdraw(account) / controller liquidity evaluated over the full enabled-collateral set.

B1.R2) Add a regression test that covers the following
- enable collateral A and B;
- create residual debt;
- fully withdraw A while B alone keeps the account healthy — must succeed;
- repeat with insufficient B — must fail through the canonical Euler health check.

B1.R3) Keep the existing full-debt cleanup path: disabling the controller and stale collateral flag after a true full close is sensible.

This is a correctness/feature blocker for #990 even though it is not a direct fund-loss bug.


## S1 — Exact repay intent can be griefed by live-debt changes

Morpho and Aave V3/V4 revert on zero debt, and an exact non-sentinel repay can fail its delta check if debt is reduced between intent creation and execution. 
A third-party repayment/donation can therefore cancel an otherwise valid intent. 
Euler is more forgiving: it caps repayment to current debt and skips when debt is already zero.

This is primarily a liveness/operations issue because the execution is atomic and Superform can simulate immediately before private submission. 

Still, the provider families should converge on one documented policy:

S1.R1) either cap exact repayment to live debt/handle zero debt gracefully across all providers; or
S1.R2) require full-repay sentinels for strategist close flows and enforce resimulation/private submission operationally.


## S2 — Refresh the stack before merge

Latest #988 is not an ancestor of current #990. The two branches independently merged newer dev after #990 was stacked. After fixing B1, merge/rebase the latest #988 head into #990 (or retarget after #988 lands), then rerun all checks. This is stack hygiene, not evidence of lost source code.



## S3 — Deployment script invokes the new loan deployment paths twice

The generic deployment run now auto-dispatches V2/Euler hooks on supported chains, while the shell orchestration also calls the explicit V2/Euler entrypoints. Deterministic deployment and output merging make this harmless, but it adds redundant Forge invocations and complicates failure diagnosis. Clean up as a small tooling follow-up.



## S4 — Standalone Euler repay discovery must be explicit

The standalone repay hook intentionally does not advertise the loan-hook interface because it has no collateral getter. Ensure OMS/Erebor discovery uses the explicit hook schema/name rather than assuming every sizing-capable loan hook is discoverable only through that ERC-165 interface.






# Production-activation gates

These are rollout gates rather than reasons to reject #988's bytecode.

## A1 — Exact leaf allowlisting is the security boundary for calldata-selected providers

For Aave V3 the pool is supplied in calldata. For Aave V4 the spoke is supplied in calldata, and reserve validation asks that same spoke to map reserve IDs to tokens. Those checks prevent inconsistent data but do not independently establish that the pool/spoke is canonical.

The inspector payloads are correctly shaped to enforce the missing trust decision off-chain:
- Aave V3: pool + loan token + collateral token + variable-rate mode;
- Aave V4: spoke + loan token + collateral token + both reserve IDs;
- Morpho: singleton + loan token + collateral token + oracle + IRM + LLTV.

Before any production root references these hooks, the leaf-generation pipeline must reject anything outside an independently maintained allowlist of:
- canonical Aave V3 pools and approved reserve pairs;
- canonical Aave V4 spokes and approved reserve-ID/token pairs;
- approved full Morpho MarketParams, not merely approved tokens or the Morpho singleton;
- approved specific Euler controller/collateral vault pairs and their oracle/risk/governance configuration. isProxy proves code lineage, not market quality.

This aligns with the Superform risk framework's treatment of integration, operational and liquidation risk and with the cross-chain design note's warning that hook-based registration is unsafe without an independent validation layer.


## A2 — Callback-capable assets/providers must be excluded or the transient settlement context must be keyed

BaseLoanHookV2 deliberately uses plain, non-context-keyed transient fields. Its own invariant assumes the account cannot re-enter the same hook during settlement through ERC-777-like token callbacks or provider callbacks. Morpho calls use empty callback data, which is good, but the assumption still belongs in the production market/asset admission gate.

Minimum rollout rule: reject callback-capable tokens and any provider configuration capable of re-entering the hook chain unless settlement storage is redesigned to be context-keyed.




# CI Issues - Independent local reproduction caveat

The local run is not used as canonical pass/fail evidence because the repository's current setup is not reproducible on this environment:
- the workflow installs pnpm@8;
- the checked-in ModuleKit/Safe7579 lockfiles are lockfile version 9, which pnpm 8 reports as incompatible and ignores;
- newer pnpm versions enforce different build-script policies;
- the recursive worktree submodule graph also exceeds Windows Git path limits.

A non-CI Foundry 1.8.1 attempt compiled and ran the Euler suite but produced 78/91 passes and 13 settlement-test failures; exact CI Foundry 1.4.4 could not be reproduced cleanly after dependency resolution drift. 
This does not override the exact-head hosted result, but it exposes a pre-existing build-reproducibility weakness worth fixing separately: use a lockfile-compatible pinned pnpm version and a frozen install in CI.

NicolaBernini (changes requested):
# Overview 
Good architecture but a major finding that needs to be addressed before moving on, see details here 
https://github.com/superform-xyz/v2-core/pull/990#issuecomment-5495410880

supervaulter:
Thanks for the deep review — addressed as follows (head `01b8704f`):

**B1 (P0) — fixed.** The `RESIDUAL_DEBT_FULL_WITHDRAW` pre-revert is removed (B1.R1): the hook no longer pre-judges health, and EVK's end-of-call account-status check over the complete enabled-collateral set is the canonical arbiter. NatSpec updated to state the multi-collateral rationale explicitly. The full-debt cleanup path (controller self-disable before withdrawal, EVC collateral release after) is unchanged (B1.R3). Three regression tests added (B1.R2): full-withdraw-with-residual-debt now builds and executes when other collaterals cover the debt, and fails atomically through the (mock-emulated) provider health check when they don't. 656 loan-hook tests green.

**S1 — acknowledged, follow-up.** Euler deliberately uses cap semantics + zero-debt skip (documented in the hook NatSpec, including the third-party-gift rationale). Converging Morpho/Aave V3/V4 onto the same policy means changing #988's already-reviewed hooks, so I'd take S1.R1 (cap-to-live-debt everywhere) as a follow-up ticket rather than widening this PR; until then S1.R2 holds operationally (pre-submission simulation + private submission).

**S2 — done.** Latest #988 head merged into this branch; all checks rerun.

**S3 — acknowledged, follow-up.** The run()+explicit-entrypoint double dispatch mirrors the pre-existing pattern (HyperCore/Firelight/Aave V3 have the same shape on dev), so I'll clean up the whole pattern in one tooling pass rather than special-casing the new families here. Deterministic CREATE2 + output merging keep it harmless meanwhile.

**S4 — confirmed.** The pricing-side descriptor codec (SUP-20799) maps hooks by explicit registry name/schema, never by ERC-165 discovery, and `EulerRepayHook`'s honest non-advertisement is documented in its NatSpec. Will make sure the Erebor registration follows the same rule.

**A1/A2 — agreed as production-activation gates**; they belong to the leaf-generation allowlist and market-admission policy work and are noted in the rollout checklist rather than this PR's bytecode.

CI reproducibility (pnpm pin/frozen install) noted as a separate infra follow-up.

supervaulter:
@NicolaBernini 

B1 (P0): dropped the RESIDUAL_DEBT_FULL_WITHDRAW pre-revert entirely. You're right, it collapsed a multi-collateral account into a single collateral. EVK's end-of-call status check runs over the full enabled set, so withdrawing 100% of collateral A with residual debt is legitimate when B carries the account. The hook no longer takes a health position of its own and leaves the provider check as the arbiter. I rewrote the NatSpec to say exactly that. The full-close cleanup (controller self-disable before withdraw, EVC collateral release after) is untouched. Also added the three regression tests from B1.R2: build and full execution succeed with residual debt when other collateral covers it, and the close reverts atomically through the mock-emulated account-status check when it doesn't. 656 loan-hook tests passing.

S2: latest #988 head is merged in and everything re-ran on top of it.

S4: confirmed on the consumer side. The pricing service's descriptor codec resolves hooks by explicit registry name and schema, never via ERC-165, so the standalone repay hook's honest non-advertisement doesn't affect discovery. Erebor registration will follow the same rule.

A1/A2: agreed, treating both as production activation gates. Leaf generation gets an independent allowlist (canonical pools and spokes, full Morpho MarketParams, approved Euler controller/collateral pairs), and callback-capable assets stay excluded at market admission until settlement storage is context-keyed. Neither changes this PR's bytecode.

The remaining items felt like scope creep for this PR, so I filed them as follow-ups instead:

- SUP-20842: converge repay semantics on cap-to-live-debt across Morpho and Aave V3/V4 (S1). Euler's graceful zero-debt handling becomes the norm. Touching those hooks here would reopen #988's reviewed surface, and until it lands, pre-submission simulation plus private submission cover the griefing window.
- SUP-20843: deduplicate the deploy dispatch between the generic run() and the per-family entrypoints (S3). The double invocation actually predates this PR (HyperCore, Firelight and Aave have the same shape on dev), so the cleanup should hit the whole pattern in one pass.
- SUP-20844: reproducible CI installs. Pin a lockfile-compatible pnpm and use frozen installs so external reviewers can rebuild the exact CI environment. That one is for your local-repro caveat.

NicolaBernini:
# Review 2 - Overview 

The source-level B1 remediation is correct: the hook no longer imposes a single-collateral assumption, and EVK’s canonical account-status check is allowed to decide whether a full withdrawal is healthy across the complete enabled-collateral set. 
The three added regression cases cover the intended accept/reject behavior.

However, the deployable bytecode committed in the PR was not regenerated. 
The repository’s normal staging/production deployment path reads the locked artifact, and that artifact is still byte-for-byte the pre-fix version containing RESIDUAL_DEBT_FULL_WITHDRAW. 
As written, a standard deployment would therefore ship the old behavior even though the Solidity source and tests show the fix.

Recommendation: request one required artifact-refresh change, preferably bundle the collateral-slot cleanup and verification cleanups below, then re-review only that narrow delta.


# Findings 

## Blocker R1 — deployed bytecode is still the pre-fix implementation

Evidence

Neither deployable artifact changed between the previously reviewed head 281af058… and the new head 6ffdb4f6…:

| Artifact | Old Git blob | Current Git blob |
|---|---|---|
| script/generated-bytecode/EulerRepayAndWithdrawHook.json |7dc2cc3cf8ac4ae8c7be13cb60f2a00cf852199b | 7dc2cc3cf8ac4ae8c7be13cb60f2a00cf852199b |
| script/locked-bytecode/EulerRepayAndWithdrawHook.json | 7dc2cc3cf8ac4ae8c7be13cb60f2a00cf852199b | 7dc2cc3cf8ac4ae8c7be13cb60f2a00cf852199b |

The selector of the removed error is: `RESIDUAL_DEBT_FULL_WITHDRAW() → 0xfc5b8cff`

That selector is still present in both current JSON artifacts.


# Required remediation

Before approval:
- Regenerate EulerRepayAndWithdrawHook bytecode from the current source.
- Commit both the generated and locked JSON artifacts.
- Re-run build, the full test suite, manifest lint, and deployment smoke.
- Recompute/check deterministic deployment addresses or manifests if the changed creation bytecode affects CREATE2-derived addresses.
- Preferably add a CI freshness check that compares freshly compiled creation bytecode with the committed generated/locked artifact for every source contract changed in a PR.

This is the only hard blocker found in the new delta.

NicolaBernini (approved):
# Overview 
OK I can see the bytecode has been regenerated here 
https://github.com/superform-xyz/v2-core/pull/990/commits/7bbc5ccaad6037ae1b88d253586abb44b259b2c7

Since that was the only blocker, so LGTM now

PR #1006, opened 2026-09-16
feat: standalone loan hooks

supervaulter:
For SUP-21021

https://linear.app/superform/issue/SUP-21021/feature-add-morpho-v2-standalone-pledge-borrow-and-release-hooks


NicolaBernini:
# Overview 

PR is good and can be approved, just a couple of non-blocking nits 



# Findings 

## Finding F1 — Low / P3: the new manifest entries falsely advertise the loan ERC-165 interface

Location: [new enrichment entries, lines 389–391](https://github.com/superform-xyz/v2-core/blob/91cacfbafe7a44f5b55062312331263678c12434/tooling/hook-enrichment.yaml#L389-L391). The generated manifest repeats the claim for all three new hooks.

The loanInterfaceHooks list causes the generator to add ISuperHookLoans to each hook’s erc165 field. However, the actual contracts return false for supportsInterface(0x0255d6e4), the Solidity-computed ISuperHookLoans interface ID. This is also the deliberate expectation in the PR’s [unit test](https://github.com/superform-xyz/v2-core/blob/91cacfbafe7a44f5b55062312331263678c12434/test/unit/hooks/loan/MorphoStandaloneLoanHooksV2.t.sol#L185-L196).

Independent reproduction: compiled and deployed the unmodified three contracts, queried the interface ID and supportsInterface, and compared the results with manifests/hooks.json. All three disagree with the manifest.

Impact: a client following the manifest’s advertised discovery behavior can reject the hook or take an unexpected fallback path when it checks the chain. The loan getters themselves exist; this is an interface-discovery metadata error, not missing loan functionality. The same inconsistency exists for older V2 hooks, but this PR adds three new incorrect entries.

Smallest fix: remove these three hooks from the list that produces the ERC-165 claim and regenerate the manifest, preserving the deliberate on-chain behavior. If advertising the interface is actually desired, implement it only in the new standalone base, update the test, and regenerate/review the new locks. Do not modify deployed shared bases just to fix this metadata.

Regression: compare the advertised interface list with actual supportsInterface results. Current manifest lint synthesizes the list and does not catch this contradiction.




## Finding F2 — Low / P3: sizing APIs accept noncanonical payloads that build and inspect reject

Location: [new standalone base inheritance](https://github.com/superform-xyz/v2-core/blob/91cacfbafe7a44f5b55062312331263678c12434/src/hooks/loan/morpho/BaseMorphoStandaloneLoanHookV2.sol#L17), which inherits [BaseLoanHook.decodeAmounts](https://github.com/superform-xyz/v2-core/blob/91cacfbafe7a44f5b55062312331263678c12434/src/hooks/loan/BaseLoanHook.sol#L39-L68)[ and ](https://github.com/superform-xyz/v2-core/blob/91cacfbafe7a44f5b55062312331263678c12434/src/hooks/loan/BaseLoanHook.sol#L39-L68)[replaceCalldataAmounts](https://github.com/superform-xyz/v2-core/blob/91cacfbafe7a44f5b55062312331263678c12434/src/hooks/loan/BaseLoanHook.sol#L39-L68) without adding the V2 decoder checks.

For all three hooks, execution and inspection call _decodeMorphoV2(data, true), but the sizing methods merely read or overwrite the primary word at byte 132.

Independent reproduction, each of three hooks:



| Input defect | decodeAmounts / replaceCalldataAmounts | build / inspect | 
|---|---|---|
| 229-byte payload | Succeeds | INVALID_DATA_LENGTH | 
| 231-byte payload | Succeeds | INVALID_DATA_LENGTH | 
| Nonzero secondary word | Succeeds | RESERVED_FIELD_NOT_ZERO | 
| Boolean byte 0x02 | Succeeds | INVALID_BOOL_VALUE | 
| Nonzero reserved byte | Succeeds | RESERVED_FIELD_NOT_ZERO | 


The experiment used primary 1, successfully replaced it with 9, and confirmed that the resulting malformed payload remains readable by the sizing API.

Impact: off-chain sizing/validation can accept and transform an invalid V2 payload that later fails inspection or execution. This is a strict-parser/acceptance-criteria gap, especially against the ticket’s “decodes/replaces only primary and rejects a nonzero secondary” requirement. It is not an execution-validation bypass: mandatory build/inspect checks still fail closed.

Fix: add standalone-base overrides that run _decodeMorphoV2(data, true) before reading/replacing primary, while preserving the one-element replacement-array check. This confines the change to the three new hooks. The inherited decodeUsePrevHookAmount already checks length and canonical boolean, but is not a complete reserved-field validator; document that boundary or make it consistent where feasible.

Regression: test malformed hook-data lengths, secondary word, boolean and reserved byte through the sizing APIs themselves, rather than only through build.




## Security and execution conclusions

PLEDGE: approve zero → approve exact amount → supply collateral with empty callback → reset allowance. It enforces exact measured collateral spend and publishes outAmount=0, with outToken=collateralToken.

BORROW: Morpho receives exact assets, zero shares, and the executing account as both borrower and receiver. Published output is the exact loan-token balance increase. No in-hook target-LTV or price-derived sizing.

RELEASE: calls only withdrawCollateral; no hidden repayment. Exact and full-collateral modes settle against the measured collateral-token increase. Morpho enforces position health. Independently exercised a healthy partial release with debt outstanding and an unhealthy full release that reverted with state unchanged.

PREV: correct token identity and nonzero/non-max predecessor output are enforced. Independently exercised PREV through full execution for all three hooks, including BORROW and RELEASE, which the new repository tests mostly cover at build level.

Identity: inspection binds immutable Morpho plus all five market parameters. Independent checks changed every field and the singleton; each changed inspection. Amounts, PREV flag and placeholder header changes did not alter it.

Accounting: all three remain NONACCOUNTING. The executor does not interpret their placeholder header as a ledger yield source.

Custody/authorization: provider calls originate from the executing account; no calldata-selected owner, receiver or external authorization grant is introduced.

Compatibility: no pre-existing runtime contract or locked-bytecode file is changed by the PR. The new standalone helper avoids changing the shared deployed bases.

Deployment: three distinct keys and artifacts are added to the existing Morpho V2 path, inheriting its configured-singleton chain gates. Empty new staging/prod address maps correctly reflect that deployment output is not yet present.

Artifact integrity: recompiling the unmodified source with solc 0.8.30, optimizer 200, Prague and metadata hash disabled reproduced all three committed creation bytecodes exactly; generated and locked artifacts also match.

These findings concern Superform’s versioned hooks over Morpho Blue, not Morpho’s separate Vaults V2 product. Protocol behavior was checked against the [official Morpho Blue source](https://github.com/morpho-org/morpho-blue/blob/8e26ca6a8dbc5089edcd67fb576248810fd2870a/src/Morpho.sol), as identified by [Morpho’s contract documentation](https://docs.morpho.org/developers/contracts/).





## Ticket clarifications and existing limitations

PLEDGE’s zero output is correct. Cosmin explicitly described the departure from the ticket’s literal “publish measured spend” wording in [Slack](https://superformlabs.slack.com/archives/D08HTSL4GM9/p1789561832418099). A spent amount is not a newly available balance; publishing it would mislead a downstream PREV consumer, particularly when unrelated wallet inventory remains. The implementation still checks the spend exactly, and matches the currently shipped repay settlement behavior. Amend the ticket wording; do not restore a positive PLEDGE output.

Empty-position RELEASE: max mode checks posted collateral and rejects zero before producing the provider execution. Exact/PREV modes do not pre-read the position; an empty position reaches Morpho and reverts there. I reproduced both paths. Thus the ticket’s unconditional “zero posted collateral reverts before the call” wording is broader than the implementation. This is a failure-location discrepancy, not a successful zero-position withdrawal or financial vulnerability. Either narrow the criterion to full mode or add a uniform precheck.

Max word plus PREV: the calldata primary word is ignored, including max, and the validated predecessor output becomes an exact release amount. Full-withdraw mode remains reachable only with PREV=false. This behavior is explicit in code/tests and follows the existing primary-override convention; distinguish it from a policy of rejecting every max+PREV encoding.

Same-address PREV chaining remains unsupported. A BORROW → same BORROW with PREV=true builds against the old output, then setExecutionContext changes that same hook’s context before preExecute reads it again. It fails with PREV_TOKEN_MISMATCH. Independent execution confirmed atomic rollback; repeated explicit-amount BORROW succeeds. This is a pre-existing lifecycle limitation newly reachable for these same-token producers. Keep it in routing rules, and use explicit amounts for repeated same-address operations. A future general fix belongs in the lifecycle design, not an unsafe local workaround.

Callback/reentrancy assumptions remain material, but the PR’s claimed exploit is not independent evidence. Shared balance snapshots are not keyed per execution context, so approved tokens/providers must not permit hostile interleaving. The checked-in security report labels a callback-based predecessor-output inflation scenario P2 and “accepted.” That is the author’s report, not an approval supplied by Nicola in this review. Moreover, its particular fee-transfer example is incomplete: [the executor writes the cached output minus fee after the transfer](https://github.com/superform-xyz/v2-core/blob/91cacfbafe7a44f5b55062312331263678c12434/src/executors/SuperExecutorBase.sol#L214-L232), overwriting a simple inflation made during that callback. I do not promote it to a confirmed finding. This observation does not prove arbitrary callback assets safe; the inherited trusted-token/provider assumption still applies.

Exact market admission, debt/risk limits and downstream family/schema registration remain separate work. In particular, do not infer that any syntactically valid Morpho market is economically approved merely because its complete tuple is inspected.

NicolaBernini (approved):
# Overview 
PR looks good, only a couple of non-blocking nits, also added some clarification in the notes, see details 
https://github.com/superform-xyz/v2-core/pull/1006#issuecomment-5697785389

PR #1009, opened 2026-09-17
feat(hooks): standardize Morpho Blue header identity across all hooks (SUP-21038)

supervaulter:
Closes [SUP-21038](https://linear.app/superform/issue/SUP-21038/feature-all-morpho-blue-hooks-yieldsource-is-morpho-marketparams-are).

## Summary

Every Morpho Blue hook now shares the **same 52-byte strategy header as the ERC-4626 hooks**:

| Offset | Field | Value |
|---|---|---|
| 0 | `yieldSourceOracleId` | Superform Morpho Blue YS oracle id |
| 32 | `yieldSource` | Morpho Blue singleton — **the call target** |

The body is a `MarketParams` **filter only** (loan, collateral, Morpho `IOracle`, IRM, LLTV). Morpho is neither a body field nor a separate inspect address.

Applies to all 8 ops: **Lend / Redeem** (V1 `MorphoLendHook`, `MorphoWithdrawHook`) and **OPEN / REPAY / CLOSE / PLEDGE / BORROW / RELEASE** (the 6 V2 hooks, incl. the SUP-21021 trio).

## What changed

- **Header decode + call-target pin.** Hooks decode `yieldSource` from the header (revert `ADDRESS_NOT_VALID` on zero) and pin it to the `morpho` immutable with a new `YIELD_SOURCE_MISMATCH` assertion on every `build` **and** `preExecute` path. Every Morpho call — approve spender, supply/borrow/repay/withdraw*, and `accrueInterest` — now targets the header-derived `yieldSource`. The immutable is retained purely as that pin.
- **`inspect()`** on every hook packs `(yieldSource, loan, collateral, oracle, irm, lltv)` — Morpho as the header yield source, not a 6th packed field. Now `pure`.
- **Both bases, once each (ticket point 1).** V2: implemented on `BaseMorphoLoanHookV2`; the 6 leaves swap targets/approve spenders to `vars.yieldSource` and drop the `placeholder0/1` NatSpec. V1: `BaseMorphoLoanHook`'s shared decoders extract the header yield source (zero → `ADDRESS_NOT_VALID`) and expose `_requireYieldSourceIsMorpho` / `_requireHeaderMorpho` + `YIELD_SOURCE_MISMATCH`; **all 7 V1 leaves** (`MorphoLendHook`, `MorphoWithdrawHook` and the 5 borrower hooks) pin the header on every `build`/`preExecute`, target it, and expose the 6-field inspect.
[...]

PR #1010, opened 2026-09-17
feat(hooks): MorphoLend/Withdraw as MONEY_MARKET (INFLOW/OUTFLOW) with per-market ledger keying (SUP-21024)

supervaulter:
## Summary

> **Based on `dev`** (#1009 / SUP-21038 merged as `c4480ce4`). Ticket items 1–5 (header names, `HookDataDecoder` decode + zero-revert, body unchanged, 6-field `inspect()`) are already on `dev`; this PR lands items 6–8 for `MorphoLendHook` / `MorphoWithdrawHook` **without touching the executors**, plus a small follow-up on the V1 repay hooks (accrue interest via the header-derived target; both re-locked).

Design (agreed with Ronny, 2026-09-17): the executor is untouched, so SuperLedger is keyed by whatever the header carries at offset 32. Morpho Blue is one singleton hosting many markets, so the singleton cannot be that key (every market's cost basis / PPS would collapse onto `0xBBBB…`, and `MorphoBlueYieldSourceOracle` could not resolve it). For these two MONEY_MARKET hooks, offset 32 therefore carries the **registry market key** — Morpho's 32-byte market `Id` truncated to 160 bits, i.e. `MorphoBlueMarketRegistry.computeMarketKey(MarketParams)`, which already is what the oracle resolves — and the **Morpho singleton is fixed per chain in the hook** (`morpho` immutable), never passed in the header. `yieldSourceOracleId` stays the Superform Morpho Blue YS oracle config id. LOAN hooks (V1 borrower + V2) are unchanged: singleton at offset 32 as per #1009.

## What changes

**Item 6 — type flip, leaf-only.** `MorphoLendHook` → `HookType.INFLOW`, `MorphoWithdrawHook` → `HookType.OUTFLOW`. `hookType` is plain storage on `BaseHook`, so a small new `BaseMorphoMoneyMarketHook` reassigns it after construction. No shared-base changes — the 5 V1 borrower hooks and the 6 V2 LOAN hooks are byte-identical to their #1009 locks and stay NONACCOUNTING.

**Item 7 — per-market ledger keying, hook-only.** `BaseMorphoMoneyMarketHook` derives the market key from the body MarketParams and pins the header to it on every `build` and `preExecute` (`MARKET_KEY_MISMATCH`), so a header can never name a different market than the one the body acts on — nor the singleton, nor a foreign address. Every Morpho call / approve spender targets the `morpho` immutable. `inspect()` packs singleton + MarketParams (the key is a pure function of those fields, so it is not packed separately). The unchanged `SuperExecutorBase._updateAccounting` then posts INFLOW/OUTFLOW against the market key, and `MorphoBlueYieldSourceOracle` prices it through the registry as today.

**INFLOW/OUTFLOW correctness.** Withdraw sets transient `usedShares` = supply shares actually burned (position diff, so by-assets and by-shares both work) and `asset` = loan token (the fee token); lend sets `asset`. A zero header oracle id reverts `ORACLE_ID_NOT_VALID` (item 2).

**Tooling.** Manifest generator understands an explicit `HookType` passed to an intermediate base (constructor-scoped regex) and keeps `subtype: LOAN`; lint allows `lend/instant → INFLOW`. *Incidental fix:* `SwapUniswapV4Hook` was unclassified because the old regex missed the `ISuperHook.` qualifier — its constructor says NONACCOUNTING and the manifest now agrees.

## Deviation from the ticket text

SUP-21024 says "do not put the truncated market key in `yieldSource`" and "SuperExecutor (or equivalent path) must post against the registry market key". Those two can't both hold with an untouched executor (the executor posts exactly what offset 32 carries). Per Ronny's call, executors stay frozen and the 20-byte market id goes at offset 32 with the singleton constant in the hook. The bundler must pack `computeMarketKey(MarketParams)` (or `registry.computeMarketKey(...)`) at offset 32 for these two hooks; SUP-21025 / SUP-21026 should register market keys rather than the singleton.

## Deploy impact

- `MorphoLendHook`, `MorphoWithdrawHook` (money-market change) and `MorphoRepayHook`, `MorphoRepayAndWithdrawHook` (accrue-interest follow-up) get new CREATE2 addresses. **No executor or ledger redeploy.**
- The flipped hooks post accounting, so each chain needs `MorphoBlueMarketRegistry` with the markets registered (IRM approved) and a `MorphoBlueYieldSourceOracle` wired into `SuperLedgerConfiguration` under the Morpho oracle id **before** roots use the new addresses — an unregistered market fails closed at accounting (`MARKET_NOT_REGISTERED`).
- Old lend/withdraw addresses remain valid for already-signed roots.

## Bytecode

Regenerated + re-locked for `MorphoLendHook` and `MorphoWithdrawHook` (`script/locked-bytecode` + `-dev`) and, in the follow-up commit, `MorphoRepayHook` and `MorphoRepayAndWithdrawHook` (`script/locked-bytecode`, same handling as #1009). Executors, the other V1 borrower hooks and the V2 hooks verified byte-identical to their locks (creation + runtime objects).

## Tests (all green)

- **Hook unit:** INFLOW/OUTFLOW types (LOAN hooks still NONACCOUNTING), ERC-165 list unchanged, header key == `registry.computeMarketKey` (+ fuzz; distinct per market; never the singleton), every Morpho call targets the singleton, `MARKET_KEY_MISMATCH` on build + preExecute for another market's key / the singleton / a foreign address, zero-oracle-id reverts on every entry point, withdraw `usedShares`/`asset` by shares and by assets, lend `asset`.
- **Shared-Morpho unit + mainnet fork (two real markets, all 8 ops):** lend/withdraw carry the market key, LOAN hooks the singleton; inspect and call-target invariants hold for both; wrong-header sweep covers both error paths.
- **E2E through the real ERC-4337 executor (mainnet fork, WBTC/USDC + wstETH/WETH):** lend posts `AccountingInflow` keyed by the market key; two markets keep distinct accumulators, cost basis, oracle decimals and PPS; withdrawing one leaves the other intact; withdraw posts `AccountingOutflow` with `usedShares` and pays the realized-profit fee in the loan token; partial withdraw by assets reduces the accumulator by the burned shares; a header naming the other market's key reverts `MARKET_KEY_MISMATCH` before any Morpho call or posting; an unregistered market reverts `MARKET_NOT_REGISTERED`; V2 LOAN hooks post nothing.
- Existing Morpho lend/withdraw suites (`MorphoLendIntegrationTest`, `MorphoBaseChainHooksFork`, `MorphoLendE2E`, `MorphoBlueMarketFork`, `MorphoBorrowerE2E`, sizing-interface suites) migrated to the market-key header and green; full unit tree green.

https://claude.ai/code/session_011QrUQT7gjosoqKGgWwgNMr


NicolaBernini:
# Overview 

Good PR, just one P2 finding 



## Snapshot and scope

[PR #1010](https://github.com/superform-xyz/v2-core/pull/1010): “feat(hooks): MorphoLend/Withdraw as MONEY_MARKET (INFLOW/OUTFLOW) with per-market ledger keying (SUP-21024)”.

Reviewed head: e9602466def300c6c1ccbd37163f16d18ef4c0c3.

Base: c4480ce4daf1d8ccfa0198f40097233a0a3e66f2 on dev, including merged PR #1009.

Two commits, 27 changed files. Head, base and OPEN status were rechecked before publishing.

Refreshed the main local core clone and created a detached review checkout. Reviewed the complete source/tooling/test diff, bytecode artifacts, BaseHook lifecycle, SuperExecutorBase, ledger/configuration, Morpho registry/oracle and the relevant current periphery Merkle consumer.

Periphery consumer snapshot: 61a83e6d64528ffb85fc75a58a3b297edc77ed62, after refreshing its local clone.

No tracked production source was edited, no deployment was performed, and no GitHub/Slack/Linear review message was posted.




# Findings 

## F1 — P2: return the market key in the MONEY_MARKET inspection payload

Locations: [MorphoLendHook.sol:132–138](https://github.com/superform-xyz/v2-core/blob/e9602466def300c6c1ccbd37163f16d18ef4c0c3/src/hooks/loan/morpho/MorphoLendHook.sol#L132) and [MorphoWithdrawHook.sol:137–143](https://github.com/superform-xyz/v2-core/blob/e9602466def300c6c1ccbd37163f16d18ef4c0c3/src/hooks/loan/morpho/MorphoWithdrawHook.sol#L137).

Both methods decode vars.marketKey, then discard it when constructing the inspector result:

```solidity
// Required by SUP-21024:
abi.encodePacked(marketKey, loanToken, collateralToken, oracle, irm, lltv)

// Current implementation:
abi.encodePacked(morpho, loanToken, collateralToken, oracle, irm, lltv)
```

The comment that the key is derivable from MarketParams is true, but does not preserve the required byte-level interface. Both encodings are 132 bytes; their first 20 bytes differ.

Trigger and impact. An operator or Erebor implementation following SUP-21024/SUP-21025 generates a leaf with the market key as the first inspected address. [SuperVaultStrategy._validateHook](https://github.com/superform-xyz/v2-periphery/blob/61a83e6d64528ffb85fc75a58a3b297edc77ed62/src/SuperVault/SuperVaultStrategy.sol#L1129) passes the hook's actual inspect() bytes to the aggregator. [SuperVaultAggregator._createLeaf](https://github.com/superform-xyz/v2-periphery/blob/61a83e6d64528ffb85fc75a58a3b297edc77ed62/src/SuperVault/SuperVaultAggregator.sol#L1341) hashes hookAddress and those bytes. It does not derive the key or normalize the payload. Consequently, the compliant off-chain leaf and the on-chain leaf differ and the corresponding authorization proof cannot validate. Consumers interpreting the first inspected field as yieldSource also receive the singleton rather than the market key.

This is a deterministic interoperability failure under the agreed schema, not a claim that the unfinished SUP-21025 implementation is already deployed.

Independent reproduction. I compiled and deployed the unchanged PR hooks to a local EVM, created two actual Morpho markets sharing one singleton and loan token, and reproduced all of the following on both hooks in both markets:

A valid market-key header passes build().

inspect() starts with the singleton, not the market key.

Changing only the nonzero header key leaves inspect() unchanged.

The required inspection encoding and the actual encoding produce different leaves using the aggregator's exact hashing formula.

For the first local lend case, the required key was 0x13061c28b51bd1fe19b0cdf5a68a22de57d9658ca, while the inspector returned singleton 0x5dddfce53ee040d9eb21afbc0ae1bb4dbb0ba643a. The expected leaf was 0x0252c76e017315b560a7587a9893468f632e393fd76af4d041fb36b6ca8a4e2a; the actual leaf was 0x4da4535efab2445750f1b0e20b206c3b95ca57fb8e3332bcff7b562c9a571902. These are synthetic local addresses, not production addresses.

Security boundary. The full MarketParams remain inspected, and the build/preExecute key-to-body check works. A substituted header still reverts before Morpho execution. I did not reproduce an unauthorized market substitution or theft. P2 reflects the integration/acceptance failure and the need to fix the format before publishing new roots.




### Why CI does not catch it. 

[The unit inspector expectation](https://github.com/superform-xyz/v2-core/blob/e9602466def300c6c1ccbd37163f16d18ef4c0c3/test/unit/hooks/loan/MorphoLoanHooks.t.sol#L306) and [the fork inspector expectation](https://github.com/superform-xyz/v2-core/blob/e9602466def300c6c1ccbd37163f16d18ef4c0c3/test/integration/morpho/MorphoHeaderIdentityFork.t.sol#L151) explicitly expect the singleton. The [SuperVault lender E2E fixture](https://github.com/superform-xyz/v2-core/blob/e9602466def300c6c1ccbd37163f16d18ef4c0c3/test/integration/morpho/MorphoLendE2E.t.sol#L126) mocks aggregator validation to return true. The real ERC-4337 accounting tests exercise a different path and do not test this market-key Merkle contract.

Recommended correction. Return vars.marketKey as the first inspection field in both hooks, retaining the same five MarketParams and 132-byte length. Keep Morpho as the immutable execution target and keep the existing key equality checks. Update NatSpec and inspector expectations, regenerate the two changed hook locks/addresses, and add a strategy authorization regression using a real proof whose leaf is encoded according to SUP-21025. Verify header-key sensitivity, each MarketParams field, and invariance when only amount fields change.

NicolaBernini (changes requested):
# Overview 
Good PR, just 1 P2 and somnething to fix in the CI, see details here 
https://github.com/superform-xyz/v2-core/pull/1010#issuecomment-5721499745

supervaulter:
Thanks — F1 addressed.

**F1 — inspect() is now market-key first.** Both `MorphoLendHook.inspect` and `MorphoWithdrawHook.inspect` return `abi.encodePacked(marketKey, loan, collateral, oracle, irm, lltv)` (132 bytes, `pure`). The Morpho singleton stays the `morpho` immutable (sole execution target / approve spender) and the build + preExecute `header == computeMarketKey(body)` pin is unchanged. NatSpec updated on both hooks and the money-market base.

**Tests.**
- Unit / shared-Morpho / two-market fork inspector expectations flipped from singleton-first to key-first.
- New `test_MoneyMarket_Inspect_KeyFirst_AggregatorLeafParity` (unit) and leaf-parity asserts in `MorphoHeaderIdentityFork`: the leaf computed with the aggregator's exact formula (`keccak256(bytes.concat(keccak256(abi.encode(hook, args))))`) over the SUP-21025 encoding equals the leaf over `inspect()`, and the singleton-first encoding yields a different leaf. Also pins header-key sensitivity (field 0 follows the header), sensitivity to each MarketParams field, and invariance when only amount / assets / shares fields change.
- New `MorphoLendStrategyProofFork` (mainnet fork, real deployed SuperUSDC strategy + aggregator, **no `validateHook` mock**): a two-leaf tree built off-hook from the SUP-21025 encoding is proposed by the strategy's real main manager, executed after the real timelock, and lend + withdraw run through `executeHooks` with real sibling proofs. A singleton-first root is rejected by `validateHook` and the strategy reverts before any Morpho call; through the real aggregator, header-key and lltv changes break the proof while amount-only changes stay authorized.

**Also in this push (from our own security pass on the PR, report in `specs/security-reports/2026-09-18-sup-21024-morpho-money-market-keying.md`):** gas cleanup (withdraw decoder takes `bytes memory` — one calldata copy instead of nine; single decode in both `_postExecute`s; unused imports dropped), the `BaseMorphoLoanHook` NatSpec now describes the money-market deviation and has correct byte counts (comment-only; the five V1 borrower hooks and the executor verified byte-identical), and three review-recommended E2E/fork tests: Lend(A)→Lend(B)→Withdraw(A) in one userOp, a two-account bundle sharing the hooks, donated (`onBehalf`) shares fee-capped to the tracked accumulator, and callback-data-empty for lend + V1 repay hooks on both real markets.

Lend/Withdraw bytecode re-locked (both lock dirs); no other artifact moved.

NicolaBernini:
# Review Round 2 - Overview 

PR is clean, just a few minor things 

T1) Docs Cleanup

T1.1) **PR Description:** 
The PR body still references the old singleton-first inspection encoding and an obsolete ticket deviation, and makes a blanket claim about V1 byte identity. 
Because `MorphoLendHook` and `MorphoWithdrawHook` were updated to the 132-byte market-key-first layout, their creation bytecode changed. 
Updating the PR description ensures the git history accurately reflects what actually landed for future auditors



T1.2) **Author Security Report (`specs/security-reports/2026-09-18-sup-21024-morpho-money-market-keying.md`):** 
The report targets the previous commit hash (`e9602466`), states that `inspect` must be `view` (it is now `pure`), and lists several tests as outstanding that you have since added in commit `15b8ff9`. Aligning the report with the current commit ensures our security documentation stays consistent with the codebase.




T2) CI Workflow False-Greens (Preventing False Confidence)

**Informational Smoke Jobs:** The treasury and deployment smoke jobs show green checkmarks, but looking at the test logs, their scripts exited prematurely due to environment argument and script path errors masked by `exit 0`. 
This isn't a defect introduced by PR #1010, but we shouldn't treat those badges as passing treasury/deployment smoke coverage. 
A separate CI follow-up ticket to expose script failures properly would be ideal




T3) Release and Integration Reminders

**Output Denomination (SUP-21005):** `MorphoLendHook` publishes its `outAmount` in supply shares rather than raw asset units. Downstream consumers (like SUP-21025) must handle this value as shares.
**Deployment Setup:** Canonical singleton/IRM selection, registry membership, and yield-source oracle configuration IDs need to be properly set during release deployment coordination with SUP-21025 and SUP-21026[4][7]






## Reviewed snapshot

[PR #1010](https://github.com/superform-xyz/v2-core/pull/1010), implementing [SUP-21024](https://linear.app/superform/issue/SUP-21024), with the inspection interface consumed by [SUP-21025](https://linear.app/superform/issue/SUP-21025).

New head: 15b8ff93857bfe2c099b5d60f97f3ee64e2d11f5.

Previous reviewed head: e9602466def300c6c1ccbd37163f16d18ef4c0c3.

Base remains c4480ce4daf1d8ccfa0198f40097233a0a3e66f2 on dev.

One new commit, 16 changed files: lender inspection/decoder cleanup, shared NatSpec, six lender bytecode artifacts, regression tests and an author security report. The earlier full review remains the baseline for unchanged code.

Read [Cosmin's response](https://github.com/superform-xyz/v2-core/pull/1010#issuecomment-5726399899), refreshed the local core clone, created a detached checkout, reviewed the complete incremental diff, and rechecked the current Linear issue and comments. The current ticket still requires market-key-first inspection.




# Findings 

## F1 — P2 inspection format mismatch: CLOSED

Both [MorphoLendHook.inspect](https://github.com/superform-xyz/v2-core/blob/15b8ff93857bfe2c099b5d60f97f3ee64e2d11f5/src/hooks/loan/morpho/MorphoLendHook.sol#L128) and [MorphoWithdrawHook.inspect](https://github.com/superform-xyz/v2-core/blob/15b8ff93857bfe2c099b5d60f97f3ee64e2d11f5/src/hooks/loan/morpho/MorphoWithdrawHook.sol#L139) now return the required 132-byte payload:

```solidity
abi.encodePacked(marketKey, loanToken, collateralToken, oracle, irm, lltv)
```

The immutable Morpho singleton remains the execution target and approval spender. Build and preExecute still recompute the key from the body and reject mismatches. Making the inspectors pure is appropriate because they no longer read the immutable.

Removing the singleton from these inspection bytes does not leave it unconstrained: the Merkle leaf also commits to the hook address, whose constructor fixes that singleton. The full MarketParams remain explicit in the payload. The inspector need not independently reject a mismatched key/body pair because the execution paths enforce that relationship before effects.

I converted the original independent reproducer into closure assertions and reran it against this exact head. On both hooks in both synthetic markets:

Actual inspection bytes equal the independently encoded market-key schema, and the exact aggregator leaf hashes match.

Singleton-first encoding produces a different leaf.

Changing only the header key changes the first inspected field and the leaf.

Changing each of the five MarketParams fields independently changes the inspection bytes and leaf, even with the header held fixed.

Changing only amount slots preserves inspection bytes, including both withdraw sizing slots.

A foreign or singleton-as-key header still fails build and preExecute; the inspection change does not weaken execution pinning.

For example, the first lend case now returns expected and actual leaf 0x0252c76e017315b560a7587a9893468f632e393fd76af4d041fb36b6ca8a4e2a. Round 1 returned 0x4da4535efab2445750f1b0e20b206c3b95ca57fb8e3332bcff7b562c9a571902 for that same local fixture. These are test-fixture hashes, not production configuration.

Review of the new regression coverage

The new [MorphoLendStrategyProofFork suite](https://github.com/superform-xyz/v2-core/blob/15b8ff93857bfe2c099b5d60f97f3ee64e2d11f5/test/integration/morpho/MorphoLendStrategyProofFork.t.sol#L57) directly addresses the earlier integration gap. It builds the required inspection encoding independently of the hooks, constructs a two-leaf tree with the aggregator's double-hash formula and sorted sibling hashing, and installs the strategy root through the deployed aggregator's manager/timelock lifecycle.

It then executes both lend and withdraw through the deployed SuperUSDC strategy using real sibling proofs and checks position/token effects. A singleton-first root fails direct aggregator validation and causes lending through the strategy to revert. A third test checks header/LLTV sensitivity and amount invariance through the real aggregator. Only governor registration of the freshly deployed hooks is mocked; neither manager authorization nor proof validation is mocked. This is materially stronger than the older lender E2E fixture that unconditionally mocked validation.

Additional tests cover [same-userOp market ordering and two-account hook reuse](https://github.com/superform-xyz/v2-core/blob/15b8ff93857bfe2c099b5d60f97f3ee64e2d11f5/test/integration/morpho/MorphoHeaderIdentityE2E.t.sol#L536), [untracked donated shares](https://github.com/superform-xyz/v2-core/blob/15b8ff93857bfe2c099b5d60f97f3ee64e2d11f5/test/integration/morpho/MorphoHeaderIdentityE2E.t.sol#L624), and [empty callback data for lend and both V1 repay hooks](https://github.com/superform-xyz/v2-core/blob/15b8ff93857bfe2c099b5d60f97f3ee64e2d11f5/test/integration/morpho/MorphoHeaderIdentityFork.t.sol#L241). The two-account test alternates userOps in one Foundry transaction, preserving transient state between operations; it is not a single multi-operation EntryPoint batch. The donation test checks the zero-accrual cap case, not an exhaustive donation/interest campaign.




## Decoder cleanup and accounting

The new postExecute paths reuse decoded fields rather than re-extracting the same token/market. Withdrawal's decoder now takes memory, consistent with its BytesLib consumers. Offsets, minimum lengths, oracle-id checks, address guards and amount XOR behavior are unchanged. Both hooks still measure actual Morpho position deltas; withdrawal still measures the actual loan-token balance increase and actual shares burned.

The independent local EVM run used the real SuperExecutor, SuperLedger, ledger configuration, Morpho registry/yield-source oracle and Morpho Blue implementation. It reconfirmed:

Separate share/cost-basis accumulators for two markets sharing the same loan token and singleton, with no accumulator under the singleton.

Correct partial withdrawal by assets, share-denominated withdrawal, fee token, and isolation of the untouched market.

Exactly 1% of realized profit collected after synthetic interest accrual: 12,976 loan-token base units in the fixture, with net hook output after the fee.

Withdrawal followed by lending with usePrevHookAmount reinvests the net asset output.

Approval reset after supply and account ownership/withdrawal receivers.

Foreign registered-key rejection and atomic rollback of an unregistered-market operation.

No executor or ledger production code changes were introduced in this increment.





## CI evidence caveat: both informational smoke jobs are green, but their scripts did not complete validation. 

The [treasury job](https://github.com/superform-xyz/v2-core/actions/runs/35317174423/job/105514551740) exits with “Missing required argument: environment” (then an undefined helper); the [deployment job](https://github.com/superform-xyz/v2-core/actions/runs/35317174423/job/105514551647) cannot find script/run/test/networks-production.sh. The workflow deliberately ends these steps with exit 0. This wiring is unchanged by PR #1010 and is not a new hook defect, but these green badges must not be counted as successful treasury/deployment smoke coverage. Correct the invocation/path and expose validation failures in a separate CI follow-up. This also qualifies round 1's shorthand statement about smoke jobs passing: job status alone is insufficient evidence. The 6,760-test Foundry result above is independently confirmed by its actual test log.

NicolaBernini (approved):
# Overview 
PR is good, no blocking findings so approved, however, looks like there are a few things to check in the CI, see details 
https://github.com/superform-xyz/v2-core/pull/1010#issuecomment-5726857657

PR #1011, opened 2026-09-18
feat: [SUP-21005] morpho lend amount roles assets

supervaulter:
## Summary

Implements [SUP-21005](https://linear.app/superform/issue/SUP-21005/feature-publish-morpholend-amountroles-as-assets-in-keep-share-delta): Morpho idle lend is published to the OMS as **MONEY_MARKET ASSETS/SHARES** — the same value-flow as ERC-4626 deposit/redeem — without changing Morpho calls or calldata layouts. Based on `dev` after #1010 (SUP-21024) merged; also lands the round-2 review follow-ups from #1010 (T1.2, T2, T3).

Only one contract changes: `MorphoLendHook` (new CREATE2 address; `MorphoWithdrawHook` and every other hook are byte-identical to `dev`).

## What changes (ticket items)

| # | Item | Change |
|---|---|---|
| 1 | `amountRoles` = one slot **IN / ASSETS** | Leaf override on `MorphoLendHook` (was inherited `IN/TOKEN` from `BaseLoanHook`, which stays TOKEN for the borrower family). `decodeAmounts` / `replaceCalldataAmounts` untouched — offset 132, one slot, bounded. |
| 2 | Keep share-delta `outAmount` | Unchanged (`_postExecute` = supply shares after − before). |
| 3 | `outToken` ≠ loan token | `outToken` = the **header market key** (`vars.marketKey`, pinned `== computeMarketKey(body)` in `_preExecute`) — the same synthetic share identity the executor posts INFLOW against; mirrors `ApproveAndDeposit4626VaultHook` (`outToken = yieldSource`). Chosen over `address(0)` because `FeeSplittingHook` / `WrappedNativeHook` treat zero as native ETH. PREV into a loan-token hop now fails the previous-output token check (`BaseLoanHookV2.PREV_TOKEN_MISMATCH`, Aerodrome `PREV_HOOK_TOKEN_MISMATCH`). |
| 4 | Manifest / enrichment | `tooling/hook-enrichment.yaml` `MorphoLendHook: [{IN, ASSETS}]` replaces the deferred-UNITS comment; `manifests/hooks.json` regenerated (`TOKEN → ASSETS` only); lint passes. |
| 5 | Addresses / calldata unchanged | Calldata layout and lengths unchanged (197 lend / 228 withdraw). `MorphoLendHook` bytecode necessarily changes → re-locked (both lock dirs); it already moves in the same deploy cycle as #1010, so no extra redeploy. |
[...]

PR #1012, opened 2026-09-18
feat(hooks): [SUP-21058] all Morpho LOAN hooks carry the market key at header offset 32

supervaulter:
## Summary

Implements [SUP-21058](https://linear.app/superform/issue/SUP-21058): every Morpho hook now carries the **registry market key** at header offset 32, so the whole 13-hook family shares one header rule and one leaf encoding.

The 11 LOAN hooks (5 V1 borrower + 6 V2) previously required offset 32 to be the **Morpho Blue singleton**, which doubled as the call target. Because the singleton is fixed per chain, that slot identified nothing — it could not distinguish one market from another inside Morpho. The money-market pair already carried the market key after #1010. This aligns the remaining 11, which is what makes the yield-source selector meaningful for indexing.

The key is `address(uint160(uint256(Id.unwrap(marketParams.id()))))`, byte-identical to `MorphoBlueMarketRegistry.computeMarketKey`. The singleton moves into the `morpho` immutable, which becomes the sole call target and approve spender.

Based on `dev` after #1009, #1010 and #1011. **No executor, ledger, registry or oracle changes.** `BaseLoanHook` / `BaseLoanHookV2` (shared with the Aave families) are untouched.

## What changes

| Area | Change |
|---|---|
| `BaseMorphoLoanHook` | Gains `_marketKey` + `_requireHeaderIsMarketKey` (both `pure`) and `MARKET_KEY_MISMATCH`, hoisted from `BaseMorphoMoneyMarketHook`. `_requireYieldSourceIsMorpho`, `_requireHeaderMorpho` and `YIELD_SOURCE_MISMATCH` are removed. |
| `BaseMorphoLoanHookV2` | Own copy of the same three (the two trees descend separately from the untouchable `BaseLoanHook` / `BaseLoanHookV2`). `_accrueInterest` now targets the immutable. |
| `BaseMorphoMoneyMarketHook` | Drops the three now-inherited members; NatSpec rescoped to what is money-market-specific. |
| 11 leaves | Every `Execution.target`, approve spender and `IMorpho*` receiver is the `morpho` immutable. Struct field `yieldSource` → `marketKey`. |
| `inspect()` | Key-first, 132 bytes, still `pure`. |
| Invariants held | LOAN hooks stay `NONACCOUNTING`. Calldata layouts, offsets and lengths unchanged (V1 197/198/230, V2 230). |

### Call-target surface

| | Before | After |
|---|---|---|
| `Execution.target` = `morpho` | 2 | **20** |
| `Execution.target` = calldata-derived | 18 | **0** |
| approve spender = `morpho` | 3 | **29** |
| approve spender = calldata-derived | 26 | **0** |
| empty Morpho callback args | 13 | 13 |

The old pin's effect (target equals the singleton) is now a *structural* property rather than a runtime assertion. A future refactor that forgets the pin can no longer redirect a call — only skip an identity check.

### Latent bug closed

The V1 repay and borrow layouts read `lltv` at different offsets (164 vs 197) and `_decodeHookData` guards length with `<` rather than `!=`. A 230-byte borrow payload fed to `MorphoRepayHook` used to pass the singleton pin and build a repay against a **wrong-`lltv` market**. It now fails `MARKET_KEY_MISMATCH`. Not the ticket's goal, but a real fix.

## Blast radius

Verified empirically by snapshotting all **554 compiled contracts** before and after: **exactly the 11 LOAN hooks move.** `MorphoLendHook` and `MorphoWithdrawHook` are **byte-identical** — solc strips the two now-unused inherited internals — so this redeploys 11, not 13.

11 bytecode locks regenerated in `script/locked-bytecode/`; all 13 verified equal to a fresh build. `locked-bytecode-dev/` deliberately untouched (the Morpho deploy path reads `locked-bytecode/` only).

## Security review

`specs/security-reports/2026-09-18-morpho-loan-hooks-market-key.md` — **PASS**, no P0/P1/P2. Three parallel reviewers (vulnerability scan against the internal database, an old-vs-new mechanical pin-coverage audit, coding standards) plus independent verification.

Confirmed:
- **Pin coverage 22/22.** The pin precedes every `Execution`, every provider read and every `prevHook` call on both entry points of all 11 leaves. All 26 guarded entry points keep exactly one guard; nothing was unpinned.
- **The three rewritten V1 preExecutes are strictly stronger** — zero headers still reject with `ADDRESS_NOT_VALID`, and length enforcement plus four body zero-checks plus the five-field pin are newly added.
- **`inspect()` is behaviourally identical.** The payload→bytes map is unchanged, so Merkle leaf hashes do not move for identical calldata. What changes is which calldata the hooks accept.
- **Collisions are negligible** and have zero on-chain impact here: the hook passes the full 256-bit `MarketParams` to Morpho, so the truncated key is never dereferenced on the LOAN path.

Ten P3 items (robustness, test coverage, docs). Four worth doing before merge, tracked below.

## Follow-ups

**Recommended before merge (P3, not blocking):**
- V2 leaves pin a throwaway `MarketParams` copy, then build a second one for use. Correct today, but `MorphoSupplyAndBorrowHookV2` already mutates `vars` in between — pin the struct you actually use.
- `MorphoRepayHookV2` and `MorphoRepayAndWithdrawHookV2` have **no negative-pin test on `_preExecute`**, and they are the only two whose preExecute makes a state-changing `accrueInterest` call.
- `MorphoLoanHooks.t.sol:2060` is vacuous — it compares `inspect()` output against its own input header, so it would pass with `_marketKey` deleted.
- `test_AllOps_InspectIsPerMarket` passes for the wrong reason; assert the first 20 bytes, which is what actually changed.

**Before deploy:** add a post-deploy `assertEq(hook.morpho(), MORPHO_BLUE)` for all 13 in the deploy script. The immutable is now the sole authority for every call target. (To be precise, the old pin did not validate it either — it checked consistency — so the risk is unchanged in kind, but the checklist item has gone from belt-and-braces to the only guard.)

**Integration contract:** SUP-21025's leaf builder now needs **one key-first encoding for all 13 hooks**, and the bundler must pack `computeMarketKey(MarketParams)` at offset 32 for every Morpho hook. SUP-21038 / SUP-21026 text still says "yieldSource = the Morpho singleton" and needs realigning.

⚠️ Any off-chain builder still emitting the singleton at offset 32 produces leaves that hash and verify fine, then revert at execution with `MARKET_KEY_MISMATCH`. Fail-closed, but the break surfaces at execution time rather than at proof construction.

## Tests

648 loan-hook unit tests pass. New acceptance coverage:
- `test_AllOps_RevertIf_HeaderIsSingleton` — the SUP-21038 reversal proof: the singleton in the header is now rejected by every op.
- `test_AllOps_RevertIf_HeaderIsOtherMarketKey` — the pin is per-market, not merely "not the singleton".
- `test_AllOps_InspectIsPerMarket` — two markets on one singleton are now distinguishable in field 0.
- `test_V1Borrowers_RevertIf_HeaderIsSingleton` — build **and** preExecute.
- New `test/utils/MorphoMarketKey.sol`, a file-level helper any suite can import; cross-checked against a live `registry.computeMarketKey`.

https://claude.ai/code/session_011QrUQT7gjosoqKGgWwgNMr


NicolaBernini:
# Overview 

The PR is good and can be approved, just a few of non blocking P3s




# Findings 

## F1 — P3 (new): script/generated-bytecode/MorphoLendHook.json is stale, and MorphoWithdrawHook.json was not regenerated

The PR commits a modified [generated-bytecode/MorphoLendHook.json](https://github.com/superform-xyz/v2-core/blob/f7b9fb1671cb57b367747a7bf6b45506ddb36e7e/script/generated-bytecode/MorphoLendHook.json), but it was not produced from the PR tree:



| Check on the committed JSON | Result | 
|---|---|
| ABI still declares `YIELD_SOURCE_MISMATCH()` | Yes — the error no longer exists in any source file on the branch | 
| `rawMetadata.sources["…/BaseMorphoLoanHook.sol"].keccak256` | `0x4ecb0ace…` — identical to the `dev` artifact, although the file changed in this PR | 
| Same field in the regenerated `MorphoBorrowHook.json` | `0xc8154d62…` — the PR tree's hash | 
| `bytecode` / `deployedBytecode` vs `dev` | Identical (consistent with the byte-identical claim) | 
| `MorphoWithdrawHook.json` | Not touched at all | 

The only source-hash that changed in the Lend JSON is MorphoLendHook.sol itself (the AMOUNT_POSITION NatSpec from #1011), i.e. the artifact was regenerated from dev/#1011 before this PR's edits, not after. Nothing on-chain is affected — bytecodeHash: none, deploy scripts read locked-bytecode/, and creation bytecode is equal — but the artifact's ABI carries a phantom error selector for any consumer that decodes reverts from these JSONs, and its metadata will not match the branch source. The PR's report ("diff confined to sourceMap and metadata (inherited NatSpec)") is therefore not accurate. Fix: run script/run/tooling/regenerate_bytecode.sh on the final tree (at least for MorphoLendHook and MorphoWithdrawHook) and commit; that regeneration is also the cheapest way to make the "13 verified equal to a fresh build" statement reproducible from the repo.





## F2 — P3 (optional hardening): run the pin inside inspect()

inspect() echoes the unvalidated header key (the PR's own P3-3), and the PR's ⚠️ warns that a builder still emitting the singleton "produces leaves that hash and verify fine, then revert at execution". Both inspect() and _requireHeaderIsMarketKey are pure, so calling the pin from inspect() costs nothing in purity or in the sizing views, keeps the payload→bytes map identical for every payload that can execute, and moves the failure to proof construction: SuperVaultStrategy._validateHook, Bank._validateHookConfiguration (which already wraps inspect() in try/catch) and the Erebor leaf builder all call inspect() before build(). It also closes the off-chain exposure of a consumer reading [0:20] from an inconsistent payload. The cost is that [test_Inspect_PacksHeaderYieldSource](https://github.com/superform-xyz/v2-core/blob/f7b9fb1671cb57b367747a7bf6b45506ddb36e7e/test/unit/hooks/loan/MorphoLoanHooksV2.t.sol#L435) flips to an expectRevert — that test today asserts that a singleton header still round-trips through inspect(), which is precisely the gap. If the team prefers to keep inspect() non-reverting (a legitimate choice: a mismatched payload yields a different leaf as well as a revert), then rename/repurpose that test so it does not read as an endorsement of the old encoding, and document the convention at the ISuperHookInspector interface as the report suggests.







## F3 — P3: test hygiene — concur with the PR's own list

I agree with P3-1 (V2 leaves pin a throwaway MarketParams copy and then rebuild one — correct today because _marketParams reads only the five identity fields and _resolveOpenAmount1 mutates amount1, but MorphoSupplyAndBorrowHookV2 [lines 68–73](https://github.com/superform-xyz/v2-core/blob/f7b9fb1671cb57b367747a7bf6b45506ddb36e7e/src/hooks/loan/morpho/MorphoSupplyAndBorrowHookV2.sol#L68-L73) is exactly the shape where a later refactor breaks it silently), P3-2 (no negative-pin preExecute test for MorphoRepayHookV2 / MorphoRepayAndWithdrawHookV2, the two whose preExecute makes a state-changing accrueInterest; test_AllOps_RevertIf_HeaderIsOtherMarketKey at line 256 covers 6 of 8 ops), P3-5 ([line 2060](https://github.com/superform-xyz/v2-core/blob/f7b9fb1671cb57b367747a7bf6b45506ddb36e7e/test/unit/hooks/loan/MorphoLoanHooks.t.sol#L2060) compares inspect() against its own input header and would pass with _marketKey deleted) and P3-6 (test_AllOps_InspectIsPerMarket only becomes non-vacuous at the bytes20 assertions). None of these hides a defect in the shipped source; they weaken regression protection for the next refactor. Cosmin's "four worth doing before merge" is the right call.






## F4 — Integration notes (off-chain, not blocking)

The Erebor side of this change is [SUP-21037](https://linear.app/superform/issue/SUP-21037) (LOAN V2 inspect params: yieldSource = market key with ref_type: yield_source, leaves = key + MarketParams), not SUP-21025, which explicitly scopes LOAN hooks out. The PR's follow-up should point at SUP-21037; SUP-21038 already matches the new behaviour and needs no realignment.

supervault-pricing (app/services/positions/codec.py, [_MORPHO_V2_SCHEMA](https://github.com/superform-xyz/supervault-pricing/blob/main/app/services/positions/codec.py) at line 97) names V2 inspect field 0 morpho, and validate_schema_against_registry (line 257) fails closed on any field-name drift against Erebor's inspect_params. Once SUP-21037 renames the field to yieldSource, Morpho LOAN pricing fails closed until the local schema is renamed and, ideally, uses the key to resolve the market. The 80-byte _MORPHO_V1_SCHEMA has been stale since #1009 (V1 inspect is 132 bytes, key-first); V1 borrowers are not in use, so this is hygiene. Worth a Superman-side ticket now rather than at rollout.

Any off-chain builder that still packs the singleton at offset 32 for the new hook addresses will produce leaves that verify and then revert MARKET_KEY_MISMATCH at execution (fail-closed; F2 would make it fail at construction). Old hook addresses keep accepting the singleton for already-signed roots.

Redeploying 11 hooks means 11 new addresses in SuperGovernor hook registration, Erebor hook rows and any bannedLeaves entries keyed by leaf hash — generic to every redeploy, listed for completeness.





## F5 — Deploy-time notes

Concur with the PR's P3-4: the morpho immutable is now the sole authority for every call target and approve allowance, guarded only by the constructor zero-check. The old pin never validated it either (it checked header-equals-immutable), so the residual risk is unchanged in kind; what is lost is the detection property that a correct payload against a misconfigured hook used to revert. Add assertEq(hook.morpho(), MORPHO_BLUE) for all 13 in the deploy verification. A wrong-chain MarketParams is still rejected by Morpho itself (MARKET_NOT_CREATED).

LOAN hooks still have no registry allowlist (NONACCOUNTING; only the money-market pair resolves through getMarketInfo). Pre-existing and documented; the key-first header is what makes a market-level allowlist possible off-chain via the SuperVault yield-source whitelist (SUP-21037).

NicolaBernini (approved):
# Overview 
PR is good and clean so can be approved, only a few non blocking P3s, see details 
https://github.com/superform-xyz/v2-core/pull/1012#issuecomment-5735104268

